PR #358, opened 2013-08-18
STAR and BY in list slices

Haspaker:
This commit enables full use of the STAR (`*`) and BY keyword in list slices. It seems a bit arbitrary to allow `*` in list indexes but not in slices, as is the case today. Similarly, `list[X to Y by Z]` only works in the current build if both X, Y and Z are integer literals, which is a bit confusing.

__Examples:__
Grab tail of list: `list[1 to *]`
Grab last half of list: `list[*/2 to *]`
Grab every second item: `list[to * by 2]`
Reverse list: `list[* to 0 by -1]`
___
Disclaimer: I am unfamiliar with the compiler and have no idea if my solution is the most elegant or correct way of solving this.


Haspaker:
I'm very sorry, I've been away from Github for a while. I'll fix it tonight!

Haspaker:
See #660

PR #624, opened 2014-12-05
Drop in Ace editor to homepage compiler

tekacs:
The current textarea on the livescript.net homepage is pretty frustrating to test snippets out in.

This request swaps out the editor for a LiveScript-highlighting Ace with keybindings.

The request's pretty easy to check, barring the commit dc4bc14942d45abcad62e3bb4240fc6d60fb31f1, which introduces the editor's distribution.


impinball:
I will also throw out there that mobile platforms generally aren't the best places to inflict keybindings in websites on. So, if you can come up with some sort of feature detection, window size testing, or something like that, and on the mobile end, fall back to the textarea, it would probably look a little better.

tekacs:
Will do, when I get a bit of spare time. :)

gkz:
@tekacs  I'd definitely be interested in having this functionality for the 1.4 release. 
Would you be able to do a couple of things first?
1. The diff says that there are over 400 new files included - would it be possible to only include the ACE files needed to run the editor on the site?
2. Just show a textarea for mobile, and for desktop only activate ACE on clicking the textarea, or on double clicking an example (loading it into the editor)

PR #637, opened 2015-01-01
Add Gitter chat badge to README

anko:
The [LiveScript Gitter chat](https://gitter.im/gkz/LiveScript) is mainly inactive, but I think it could be great if we started using it. Since it allows github login, it would make it much easier to find other LiveScripters, and could reduce the multitude of [gh issues](https://github.com/gkz/LiveScript/issues) that are about really small problems.

I'm aware of [/r/livescript](http://www.reddit.com/r/livescript/) and [the Google Group](https://groups.google.com/forum/#!forum/livescript), but they're better oriented for large topics, don't work with a github account, and aren't "officially" endorsed either as far as I can tell.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,6 @@ is a language which compiles to JavaScript. It has a straightforward mapping to 
 
 Check out **[livescript.net](http://livescript.net)** for more information, examples, usage, and a language reference.
 
-### Build Status
 [![Build Status](https://travis-ci.org/gkz/LiveScript.png?branch=master)](https://travis-ci.org/gkz/LiveScript)
 
 ### Install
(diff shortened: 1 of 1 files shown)


vendethiel:
I don't know. I prefer IRC by far. @gkz ?

quarterto:
There is a [Gitter-to-IRC bridge](https://irc.gitter.im/).

anko:
I forgot we had an IRC! How active is it?

Either way we decide, I think we need to reduce the friction for getting to another human. Maybe mention the subreddit, IRC and Google Group straight in the readme, or next to the other social badges on livescript.net.

At the moment it's just Twitter and Github:

shot

anko:
@raine Great!

I see no need to duplicate the functionality that IRC clearly already provides. Closing.

gabeio:
why not at least add irc to the readme...

Issue #639, opened 2015-01-06
Conditions in object literals are duplicated in the compiled output


impinball:
## Platform

Ubuntu GNOME 14.04 64-bit

## Steps to reproduce

1. Type any object comprehension with a condition, and either return it or assign it to a variable.
2. Compile the LiveScript code.
3. Observe the compiled JavaScript code.

## Example code

This is used for the output below.

```ls
ret = {[k, v] for k, v of o | cond}
ret = {[k, v] for k, v of o | foo}
ret = {[k, v] for k, v of 0 | v != void}
ret = {[k, v] for k, v of 0 | check v}
ret = {[k, v] for k, v of 0 | check k}
ret = {[k, v] for k, v of 0 | check k, v}
```

## Expected

The condition only appears once.

```js
// Generated by LiveScript 1.3.1
(function(){
  var ret, res$, k, ref$, v;
  res$ = {};
  for (k in ref$ = o) {
    v = ref$[k];
    if (cond) {
      res$[k] = v;
    }
  }
  ret = res$;
  res$ = {};
  for (k in ref$ = o) {
    v = ref$[k];
    if (foo) {
      res$[k] = v;
    }
  }
  ret = res$;
  res$ = {};
  for (k in ref$ = o) {
    v = ref$[k];
    if (v !== void 8) {
      res$[k] = v;
    }
  }
  ret = res$;
  res$ = {};
  for (k in ref$ = o) {
    v = ref$[k];
    if (check(v)) {
      res$[k] = v;
    }
  }
  ret = res$;
  res$ = {};
  for (k in ref$ = o) {
    v = ref$[k];
    if (check(k)) {
      res$[k] = v;
    }
  }
  ret = res$;
  res$ = {};
  for (k in ref$ = o) {
    v = ref$[k];
    if (check(k, v)) {
      res$[k] = v;
    }
  }
  ret = res$;
}).call(this);
```

## Found

The condition is duplicated in each one.

```js
// Generated by LiveScript 1.3.1
(function(){
  var ret, res$, k, ref$, v;
  res$ = {};
  for (k in ref$ = o) {
    v = ref$[k];
if (cond) {
      if (cond) {
        res$[k] = v;
      }
    }
  }
  ret = res$;
  res$ = {};
  for (k in ref$ = o) {
    v = ref$[k];
if (foo) {
      if (foo) {
        res$[k] = v;
      }
    }
  }
  ret = res$;
  res$ = {};
  for (k in ref$ = o) {
[...]

PR #641, opened 2015-01-08
Add failing test for object comprehensions with guard

impinball:
Relevant comment: https://github.com/gkz/LiveScript/issues/639#issuecomment-69053238

I personally don't know enough about the inner workings of the compiler to propose any significant fix, though. (Sorry...)


impinball:
@gkz You're welcome. ;)

gkz:
I feel a better failing test would not be the compilation, but exploiting some fact that the if test is evaluated twice. Maybe if the test is a function call modifying some state

impinball:
What about this?

gkz:
Thanks!

impinball:
Welcome!

PR #645, opened 2015-01-13
Add programmatic API documentation & some syntax highlighting

impinball:
Added documentation to the best of my ability of the LiveScript programmatic API. I wouldn't be surprised if I missed a few things.

Brief description of other changes:

1. My editor automatically removes trailing whitespace. That accounts for the majority of changed (not added/removed) lines.
2. I corrected a few of the categories under "Usage" for clarity.
3. I added syntax highlighting to a single `<pre>` containing HTML, starting at line 259.

PR #646, opened 2015-01-15
Fix type annotation format

impinball:
As suggested by @vendethiel in [this comment](https://github.com/gkz/LiveScript/pull/645#issuecomment-69729187). I feel it does look better, too.


vendethiel:
Thanks for putting up another PR!

PR #650, opened 2015-01-18
fix: Compiler full screen for 1550px doesn't work.

ysmood:
My macbook's resolution is 1440x900, If you resize the browser to that size, the compiler won't go full screen.


gabeio:
I have a 17" mbp(1920x1200) it full screens perfectly...? What browser are you using?

ysmood:
@gabeio 1550px is how the code was written. 1440

PR #660, opened 2015-02-05
Enable full use of the BY keyword in list slices

Haspaker:
This commit enables full use of the `BY` keyword in list slices. Currently, `list[X to Y by Z]` only works in the current build if both `X`, `Y` and `Z` are integer literals, which is a bit confusing. E.g. `list[1 to 10 by 2]` will work while `list[1 to x by 2]` raises an error. This commit solves that.

For more information, see previous pull request #358.

Example:
```
[x, y, z] = [1 10 2]
list = [1 to 10]
list[x to y by z] #=> [ 2, 4, 6, 8, 10 ] 
```


Haspaker:
The failed test is merely the one added in #641. All others work.

src/ast.ls
```diff
@@ -2190,6 +2190,26 @@ class exports.For extends While
     head += \\n + @tab if (@item or (@index and not @object)) and \} is body.charAt 0
     head + body
 
+#### Step slice
+# Slices a list in steps
+# Makes it possible to combine non-literals and the BY keyword in slices 
+# E.g. list[1 to 10][f() to x by (1+1)]
+class exports.StepSlice extends For
+
+  makeReturn: ->
```
vendethiel: I'd rather have `(@makeReturnArg) ->`

Haspaker: Sure!

vendethiel:
@gkz?

gkz:
Cool, thanks!

Issue #452, opened 2014-01-06
Source maps?


drozzy:
Forgive me if this is a repeat but I didn't find any issues about this.

Does LiveScript compiler support generating source maps?
The reason I ask is because both coffeescript and TypeScript seem to [support](http://net.tutsplus.com/tutorials/tools-and-tips/source-maps-101/) it.

fuzzthink:
+1

demux:
+1

modo-lv:
+1

vendethiel:
"+1"s are not helping, really.

blvz:
It's hard to express how "+1"s are a annoying and disrespectful, without sounding rude.
GitHub already have a "+1 button". It's on the right side of the Star/Unstar button.

Diggsey:
Well with the lack of a voting system for issues, there's no other way to express community interest in a particular issue. Forking is not a solution.

Regardless, I'm going to take a stab at implementing this, since it's about time LiveScript got sourcemaps...

vendethiel:
If there's no developer interest in a specific feature, then community interest is a bit moot.

seanstrom:
I believe this feature has been road mapped already. #260 
I think part of the hold up is that LiveScript is/was meant to be re-written on top of CoffeeScript-Redux compiler which has source map support already. If anyone is interested in helping with source maps, looking into that would be helpful.

vendethiel:
> f CoffeeScript-Redux compiler which has source map support already

If you have the energy to work on both, do it

seanstrom:
I thinks thats fair to feel that way, and I hear where you're coming from.
So what would be the best way for me to help the project?

We all keep asking for this feature, but there's all this underlining work that needs to be done before that can happen. What areas would ideally be addressed first before Source Map support?

I would be willing to work on various things if it meant we would eventually be able to add this.

am11:
My two cents here: Sweet.js is another superset language of JavaScript, which makes use of an npm module called `source-map`, which has all the mechanics to generate source-maps. If you want to consume to keep concerns separated, you would probably just need to hook it with your complier internals.

Sweet.js: https://github.com/mozilla/sweet.js.
source-map: https://github.com/mozilla/source-map.

vendethiel:
#481

gkz:
I don't mind the `+1`s.
[...]

PR #675, opened 2015-02-23
Initial source-map support

Diggsey:
This is still quite rough around the edges and could do with a lot more testing, but it works.

Pass "-m/--map" to lsc when compiling to generate source-maps.

See #452


vendethiel:
Wow, very impressive work, nicely done :-).

Diggsey:
Currently there's no way to specify the output filename when compiling and running livescript on-the-fly, so the sourcemap you get back won't work properly. For now it will only work when compiling livescript to .js files ahead of time. However, the sourcemaps for those .js files should work just fine in the browser if they are being served as static assets.

Diggsey:
Having looked into it, it's seems that a trival change to the "liveify" plugin for browserify would enable this to work with browserify (as per coffeeify):
https://github.com/jnordberg/coffeeify/blob/master/index.js#L64

For the browser, coffeescript seems to use data-uris to specify the sourcemap, but I haven't figured out how it specifies the original source yet (or if it even does).

blvz:
I've also quickly tried to make it work with browserify with no luck. Then, I've tried to see if browserify could also bundle all generated maps, with no luck. I'll have to try again with a little more time.

Diggsey:
Source maps should work in the browser now with no extra work:
```html
<html>
<head>
</head>
<body>
<script type="text/javascript" src="browser/livescript.js"></script>
<script type="text/ls">
	debugger
	a = [1,2,3]
	b = [x*2 for x in a]
	if b[1] > 2
		console.log "Hello, world!"
</script>
<script type="text/javascript">
	require("LiveScript").go();
</script>
</body>
</html>
```

Diggsey:
@blvz I've improved the command line options, so you can now pass "--map=embedded" which will embed both the source map and the original livescript inside the generated javascript file. This is the only form which browserify will accept AFAIK.

If you use that option, and also pass "--debug" to browserify, it will attempt to generate a combined source-map for the entire set of inputs. However, in my tests it's been generating invalid source-maps as output, and I have yet to figure out why.

blvz:
Very nice. Thanks for your effort. I've test it against a front-end app that uses mithril with some views deeply nested. Indeed, sourcemaps generated with browserify are slightly incorrect.

On Chrome Dev Tools, setting breakpoints in `some-file.ls` doesn't work, but setting them in `bundle.js` do. Of course, the breakpoint now points to the slightly wrong place in `some-file.ls`.

Enabling "pause on exceptions" made Dev Tools crash, but disabling it made the exception get printed in the console with correct file name and line number. Clicking on those, I could navigate to it and set a breakpoint, pausing the exception accordingly.

In the sample below, the correct file name and line number are shown in console, but navigating to it and setting a breakpoint fails.

```LiveScript
...
  onclick: ->
    return if console.log \something # setting a breakpoint on this line is useless
...
```

gkz:
Great work!
As soon as the kinks are worked out I am ready to merge this in!

Diggsey:
At least part of the problem is due to the very old version of jison being used - I've been trying to update to a newer version, but that's proving difficult, as the jison project doesn't appear to maintain any release notes or even decent documentation, and LiveScript does some unusual things such as replacing the lexer and applying some regexes to the generated code, so would appreciate some help with that side of things (even just an explanation of what it's doing!).

The old version of jison causes two problems:
1) Multi-token grammar rules will pick up the line/column information for the last token in the rule, rather than the first
2) Column information is not tracked at all, so I had to hack that in by abusing one of the other fields

Additionally, my changes to the custom lexer in LiveScript to get it to track column information are still imperfect, although even the original line tracking was not perfect either. The problem is that they're tracked independently from the input source, and so it's very easy for them to get out of sync. Ideally it would be rewritten so that location information and input source are kept in lock-step at all times.

Once the information produced by the lexer is accurate and jison is updated, then the only possible source of errors will be within the source-node tree constructed by the compiler, and that can be improved incrementally over time, on a case-by-case basis, as it will for the most part be subjective, and is easy to change.

vendethiel:
I think coffee updated the json version. That might prove interesting to check out.

Diggsey:
Turned out that half the problem was that I'd updated to a buggy version of jison. With that solved, the update was successful, and the source-maps are MUCH better.

Diggsey:
Sourcemap generation is pretty much perfect now (within the limits of v3 sourcemaps) at least for the common AST nodes.

gkz:
Awesome, great work Diggsey!

raine:
Can this be used to get stack traces with .ls source directly instead of having to compile and then check
the line numbers?

src/lexer.ls
```diff
@@ -292,9 +304,10 @@ exports import
              then @doHeredoc code, index, q
              else @strnum q+q; 2
     if q is \"
+      debugger
```
vendethiel: stray debugger here :)

Diggsey:
@vendethiel Fixed!
@raine Yes. If you are running in a browser this should work off the bat, assuming you have sourcemaps enabled in the browser config. For node, you'll have to install the [source-map-support](https://github.com/evanw/node-source-map-support) npm package. In both cases you must tell lsc to generate source-maps using "--map=linked/linked-src/embedded".

igl:
Awesome! Always bet on LS. :)

I am going to try this out with node-source-map-support tonight.

raine:
Thank you @Diggsey, got the sourcemaps working and they seem accurate! Great work.

Here's [a simple project](https://github.com/raine/livescript-cli-boilerplate) with sourcemaps enabled if someone's interested.

Diggsey:
@gkz Anything else need doing?

gkz:
I just need to check it over and merge it in

impinball:
I'm already excited.

gkz:
Sorry guys, I've been busy with work and school. I'll take a look at this in the coming week.

gkz:
Works great! Awesome work @Diggsey!

vendethiel:
:cake:

awkay:
This officially improves my chances of getting it adopted where I work! Thanks!

blvz:
I didn't care much about it at first, but this is opening lots of opportunities, like code coverage:
https://github.com/gotwarlost/istanbul/issues/212

So thanks again, @Diggsey.

gkz:
Hi @Diggsey, I was wondering if you could give some comment on how the new source map functionality works. Ie. if I want to create a new ast node, what do I need to do? I see `sn(this, ....args)` used, anything special about how the args are laid out?

Diggsey:
"sn(this, ...args)" creates a new sourcemap node: it takes the line and column information from 'this' (an AST node), and the text content for the node is given by recursively concatenating the content of sub-nodes passed in as '...args' (with plain js strings being the leaf nodes)

The aim when generating sourcenodes is to associate the location information of an AST node with precisely the javascript which that AST node generates, and it's a little more tricky than it sounds:

When generating source-nodes for the expression '1 + 2' in some simple language, it might involve two types of AST nodes, Literal and Addition. Literal is easy, it just returns "sn(this, value)", whereas Addition is more complicated: "sn(null, left, sn(this, ' + '), right)". The "null" means the sourcenode is just used for grouping - it doesn't supply any location information.

You might be tempted to just do this: "sn(this, left, ' + ', right)", and that's what I started out doing, but the problem is that "left" or "right" might contain code which is not associated with any location in the input - that code would get erroniously associated with the location information from 'this'.

gkz:
Thanks for the info!

Would you be able to give a brief description of what the different options do (linked, linked-src, embedded, debug) to add to the docs in for the upcoming release?

Diggsey:
Sure, there are three files involved when generating sourcemaps:
a) The original LiveScript source
b) The sourcemap
c) The generated javascript source

'a' can optionally be embedded inside 'b', and 'b' can optionally be embedded inside 'c' via a comment.

`linked`: No embedding, 'c' links to 'b' via a relative path, and same for 'b' to 'a'
`linked-src`: 'b' is embedded inside 'c', but 'a' is linked to
`embedded`: Everything is embedded inside 'a'
`debug`: Same as `linked`, but will also output a human readable representation of the source-node tree (similar to the output from the 'ast' option) to a '.map.debug' file.

Use `linked` or `linked-src` if you are directly serving the output from lsc - (ie. not performing further processing). They keep the original source separate, so the javascript file is still small. `linked-src` just means you have one fewer file to carry around at the expense of increasing the size of the javascript file slightly.

Use `embedded` for everything else - it's self-contained and it's the only form most other tools such as browserify will accept as input. The file will be significantly larger, but you can remedy this by running a separate tool at the end of your build pipeline to split the output back into the 'linked' form.

impinball:
You can open a PR in the `gh-pages` branch. That's where this information is most useful.

gkz:
I'll add the information to gh-pages, but I would appreciate if people who are more familiar with source-maps than me could add some tests using the `--map` option to `https://github.com/gkz/LiveScript/blob/master/test/cli.ls`

PR #676, opened 2015-02-25
options: add alias -k for --const

raine:
-k is mentioned as an alias for `--const` in the documentation


impinball:
@vendethiel PTAL. I don't see anything wrong with it.

PR #686, opened 2015-03-16
Few more failing object comprehension tests

impinball:
gkz/LiveScript#639 still isn't fixed...here's a few more failing tests. :(

@gkz PTAL

The test simply throws at the erroneously added `.push`, which incidentally, the method's argument is what's *supposed* to be the whole statement.


gkz:
Thanks for the tests!

PR #695, opened 2015-03-27
Persist REPL history

raine:
Makes LiveScript REPL history persist across sessions. Writes to `$HOME/.lsc_history`.

I did not implement any upper limit on the history file size after which the history would be truncated. Will do that if necessary.

New to this project so let me know if the changes don't follow this project's general coding style.

Cheers.


gkz:
I'm a bit weary of writing a file to a user's home directory by default.
Certainly, there would need to be a limit to how many lines it can hold.

raine:
I share that concern as well, as much as I'd like this to be default because it's really useful.

Do you think this should be enabled with an environment variable like `LSC_HISTORY=1` or a flag in `lsc`? The latter would require history users to use an alias with `lsc`.

gkz:
Maybe perhaps: if the file exists, use it and write to it, and if it doesn't, don't create it. This way a user would simply have to `touch .lsc_history`

heavyk:
+1 on that idea ... 

obviously if the file exceeds 1MB or something, it should be truncated, yeah

raine:
What do you think is a good amount of history to have and that shouldn't be exceeded?

gkz:
500?

raine:
OK, updated to save history only if `~/.lsc_history` exists and to retain max 500 lines.

Because of the max line restriction, I changed the implementation to write history to file on exit instead on appending to file descriptor on each line entered. It was easier to do that way.

phanimahesh:
Most shells by default also add to history files on exit, so that should&trade; be okay.

raine:
Personally I'd opt to write history by default now that there's the max line limit. Having to touch the file manually makes the feature hard to discover, but I'm happy to go with it as it is.

gkz:
What if I add an option to the command line, `--toggle-history` or something like that, that will add or remove the file. So instead of touching it, you can just use the option once and then it will start recording history (and use the option again to remove the file). Your code wouldn't need to change, it would still check if it's there before writing.

blvz:
`lsc --config history=true` :)

raine:
Sounds reasonable. That or maybe `--repl-history` and `--no-repl-history`.

gkz:
I like `--repl-history` and `--no-repl-history`

phanimahesh:
What should be the default behavior? I suggest the following:

1. When history file doesn't exist
    1. Don't write history by default.
    2. Create the file at start when `--repl-history` is passed, write to it on exit. This allows you to print a warning if file can't be created.
2. When history file exists
    1. Honor flags. If `--no-repl-history` is passed, don't read/write history.
    2. If no flag is specified or `--repl-history` is specified, read history on start and write on exit.

gkz:
should using `--no-repl-history` delete the history file, or simply disable the history for that session?

raine:
Not sure if disabling history for a particular session is a very likely use case.

gkz:
Yes, that's why I think it should delete the file. Perhaps a warning and a (Y/n) before it deletes?

raine:
Is there anything I can do to help with this?

raine:
I'd really like to see this in.

auvipy:
@gkz will this be in?

src/command.ls
```diff
@@ -246,6 +250,8 @@ function output-filename filename, json
 # - __^C__: Cancel input if any. Quit otherwise.
 # - __??__: <https://github.com/joyent/node/blob/master/lib/readline.js>
 !function repl
+  MAX-HISTORY-SIZE = 500
```
vendethiel: maybe we want to move that up and use `const` just to signify it's a constant. LGTM otherwise

gkz:
thanks

Issue #298, opened 2013-04-13
lsc -ej improvements


clkao:
Currently `lsc -ej` evaluates command line argument and prints the final value in json.

Per my discussion with @audreyt I'd like to propose for `lsc -ej` to also read from stdin, much like what `lsc -j` does, and put the value into @ or it for the `-e` evaluation to use.

This will make lsc quite useful as a cli tool for processing json files.

Currently you can do:

    echo '{"x":1}' | xargs echo global.it= | lsc -r /dev/stdin -pe 'it.x'

This should be simplified as:

    echo '{"x":1} | lsc -ej '@x'

PR #706, opened 2015-04-11
JSON Improvements

blvz:
It's now possible to generate JSON with `LiveScript.compile(strSource, {json: true})`.

Also, running `lsc -a file.json.ls` won't print the AST in JSON anymore. Of course you can get it by specifying the flag `lsc -ja file.json.ls`.

And as requested in #298, we can now do:
```
$ lsc -sje '[ x * 2 for x in @xs ]' <<< '{"xs":[1,2,3]}'
[
  2,
  4,
  6
]

$ lsc -sjde 'fold1 (+), @xs' <<< '{"xs":[1,2,3]}'
6
```


src/index.ls
```diff
@@ -60,9 +66,6 @@ exports <<<
   # Same as `tokens`, except that this skips rewriting.
   lex: -> lexer.lex it, {+raw}
 
-  # Runs LiveScript code directly.
-  run: (code, options) -> do Function exports.compile code, {...options, +bare}
-
```
gkz: Why was `run` removed?

gkz:
Cool!

src/index.ls
blvz: Because it was being overwritten, before being called. So I removed it to get your attention, it worked. 😄
Any hints on how to improve it? It's being overwritten by `lib/node`.

gkz: It isn't always overridden, eg. in a browser environment.

blvz: :facepalm: sorry. It's back now.

gkz:
Great work, thanks!

gkz:
Unfortunately, this broke running code with `lsc` - require does not work properly.
Code must be run with `LiveScript.run t.output.toString!, options, true`
Note that this doesn't use `run` from `index.ls`, but `run` from `node.ls`
I haven't figured out how to be able to set `this` when using `main._compile`

gkz:
I may have come up with an ugly solution... must test it out:
```
diff --git a/src/command.ls b/src/command.ls
index 279a4a6..d59a145 100644
--- a/src/command.ls
+++ b/src/command.ls
@@ -123,11 +123,10 @@ switch
       print = json or o.print
       t.output = LiveScript.compile t.input, {...options, +bare, run, print}
       LiveScript.emit 'run' t
-      o.run-context ?= {}
-      t.result = (Function t.output.toString()).call o.run-context
+      t.result = LiveScript.run t.output.toString!, options, true, o.run-context
       switch
       | json  => process.stdout.write JSON.stringify(t.result, null 2) + '\n'
-      | o.print => console.log t.result
+      | o.print => say t.result
       throw
 
     LiveScript.emit 'compile' t
diff --git a/src/node.ls b/src/node.ls
index 7fe1bda..1258409 100644
--- a/src/node.ls
+++ b/src/node.ls
@@ -6,7 +6,7 @@
 module.exports = !(LiveScript) ->
   require! [fs, path]
 
-  LiveScript.run = (code, {filename}:options?, js) ->
+  LiveScript.run = (code, {filename}:options?, js, context) ->
     {main} = require
     # Hack for relative `require`.
     if filename
@@ -17,6 +17,9 @@ module.exports = !(LiveScript) ->
     main.paths = main.constructor._nodeModulePaths dirname
     main <<< {filename}
     js or code = LiveScript.compile code, {...options, +bare}
+    if context
+        global.__runContext = context
+        code = "return (function() {\n#code\n}).call(global.__runContext);"
     try main._compile code, filename catch then throw hackTrace e, code, filename
```

blvz:
Good. I'll take a look tomorrow, starting by writing tests for CLI.

PR #716, opened 2015-04-28
Suggestion: Just use `from` as an alias for `for from`

Haspaker:
Using `for from 1 to 10 => dothings()` is the current method of repeating a loop 10 times, and it works like it should.

However, I think the `for from` keyword is very ugly and unintuitive. You have to combine two keywords to achieve a single functionality, and grammatically it looks very weird: "I want you to count *for from* 1 to 10, please". 

It would look much better if you could just drop the unnecessary keyword and just write `from 1 to 10 => dothings()`. It also is more intuitive.  

Examples:

```
from 1 to 10
    console.log \test
```

```
random-values = [ Math.random() from 1 to 3 ]

#=> [ 0.14552273, 0.13616350, 0.055535889 ]
```

Complexity wise it is not a big deal, since it only requires adding two extra lines to the compiler.

This requires adding `from` as a reserved word, but I think that is reasonable. Assigning to `from` as a variable seems like bad practice.


Haspaker:
I'm aware that you can use `for [x to y]` as well, but I don't think it is a good alternative. The JavaScript output is a lot less optimised and harder to read (and aesthetically I don't like the unnecessary array declaration).

Compare the output of `for [x to y] => f()`

```
var i$, x$, ref$, len$;
  for (i$ = 0, len$ = (ref$ = (fn$())).length; i$ < len$; ++i$) {
    x$ = ref$[i$];
    f();
  }
  function fn$(){
    var i$, to$, results$ = [];
    for (i$ = x, to$ = y; i$ <= to$; ++i$) {
      results$.push(i$);
    }
    return results$;
  }
```

with the output from `from x to y => f()`

```
var i$, to$;
  for (i$ = x, to$ = y; i$ <= to$; ++i$) {
    f();
  }
```

vendethiel:
Seems like unnecessary complexity to me here?

Haspaker:
> Seems like unnecessary complexity to me here?

You mean in the codebase?

vendethiel:
Yes. I'm always afraid adding a more complex grammar just to save 4 characters **at some point** is not the best trade we can make :).

gkz:
Actually the diff is quite small, the only thing I'm worried about is that it's technically a backwards incompatible change

raine:
I don't know LiveScript's philosophy that well and I'm not saying this change is bad but I find that in general there are too many ways to do things and the syntax is quite forgiving in some cases, making it feel sort of arbitrary (vestiges of coco, perhaps...).

Haspaker:
@vendethiel: It's not about saving characters, but effectively communicating the intent of the code without a lot of distracting/confusing noise due to weird syntax. But I get your point :laughing: 

@gkz: Yeah, that's a problem. On the other hand, backwards incompatible changes have been introduced before (removed operators and such), so it's not unheard of. But I understand it might not be worth it for such a insignificant change. I wonder if people often assign to `from` as a variable. In that case it might be better to postpone the change to some bigger, backwards incompatible version.

@raine: 

I actually agree with you. Multiple sort-of-similar syntaxes to do exactly the same thing is a bad thing.

But I think the solution in this case is to make `for from x to y` invalid syntax in the future and only allow `from x to y`. Keep the `for` keyword for when you actually want to *capture*  a value, like `for index from x to y`. Maybe this could happen in LiveScript 2.0 or something.

`from` repeats something, and `for` captures the repeating value if you need it. 

This feels less arbitrary then requiring some parts of the expression to be left empty. Code like `for from to 10 then f()` feels a bit weird.

gkz:
I'm going to say no to this for now, as it breaks backwards compatibility, but it will be revisited if we come out with a backwards-incompatible release.

PR #735, opened 2015-05-27
Fix content and compiler width

e-jigsaw:
OSX 10.10.2 / Chrome 43.0.2357.81

I fixed compiler and content pane width because compiler pane is overlap in content pane.

Preview:

[Review on Reviewable](https://reviewable.io/reviews/gkz/livescript/735)

### DIFF
--- a/index.html
+++ b/index.html
@@ -64,7 +64,7 @@
         <li><a href="#changes">Changes from Coco</a></li>
       </ul>
     </div>
-    <div class="span4 compiler">
+    <div class="span3 compiler">
       <div class="compiler-box">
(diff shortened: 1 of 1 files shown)


gabeio:
:+1:

phanimahesh:
IMO, a better fix would be to change styles.css. There are a couple of media rules that change `.span6.content`'s size to exceed that of `.span6`. No idea who added it or why, but it doesn't make sense to me.

gabeio:
I think this was fixed with  7e93c32dd861d085248e884bec067dfa0833248a

PR #757, opened 2015-07-23
Fix to SPDX license spec

RobLoach:
`npm` [updated to use SPDX "license" metadata](https://github.com/npm/npm/releases/tag/v2.10.0). It gives you warnings if you don't use it. This PR fixes that.


gkz:
thanks

PR #786, opened 2015-11-10
Ability show sources via node-inspector

darrrk:


vendethiel:
...that doesn't show the source map though, does it?!

darrrk:
This pr add ability to require *.ls with sourcemaps for correct representation in node-inspector or maybe other debugging tools.

gkz:
thanks

PR #802, opened 2015-12-29
Wrap match-cases in parens

### LINKED ISSUE #584: pattern-match and, and, or
Simple match test with expression in case:

    match 1, 3, 3
    | 1, 1, 2 or 3 => ok 0
    | 1, 2 or 3, 3 => ok 1
    | _            => ok 0

Compiles to:

    // Generated by LiveScript 1.3.1
    var ref$;
    switch (ref$ = [1, 3, 3], false) {
    case !(1 === ref$[0] && 1 === ref$[1] && 2 === ref$[2] || 3 === ref$[2]):
      ok(0);
      break;
    case !(1 === ref$[0] && 2 === ref$[1] || 3 === ref$[1] && 3 === ref$[2]):
      ok(1);
      break;
    default:
      ok(0);
    }

The problem in words: Every case-expression is bundled in one long expression.
It should wrap braces around each column so the precedence for each case is preserved.

    1 === ref$[1] && 2 === ref$[2] || 3 === ref$[2]
    ^---- 1 ----^ && ^------------ 2 -------------^ # yay
    ^------------ 1 -------------^ || ^---- 2 ----^ # nay

Correct output:

    (1 === ref$[1]) && (2 === ref$[2] || 3 === ref$[2])

igl:
This fixes #584 and makes `match` a quite powerful pattern-matching tool.
Test included.

Compiles:

    match 1, 3, 3
    | 1, 1, 2 or 3 => ok 0
    | 1, 2 or 3, 3 => ok 1
    | _            => ok 0

To:

    // Generated by LiveScript 1.4.0
    (function(){
      var ref$;
      switch (ref$ = [1, 3, 3], false) {
      case !(1 === ref$[0] && 1 === ref$[1] && (2 === ref$[2] || 3 === ref$[2])):
        ok(0);
        break;
      case !(1 === ref$[0] && (2 === ref$[1] || 3 === ref$[1]) && 3 === ref$[2]):
        ok(1);
        break;
      default:
        ok(0);
      }
    }).call(this);

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,3 +18,4 @@ After, run `lsc -h` for more information.
 ### Community
 
 If you'd like to chat, drop by [#livescript](irc://irc.freenode.net/livescript) on Freenode IRC.
+If you don't have IRC client you can use Freenode [webchat](https://webchat.freenode.net/?channels=#livescript).


vendethiel:
Seems like something is fishy somewhere else, [because it should already be wrapped in parens...](https://github.com/igl/LiveScript/blob/master/src/ast.ls#L2532)

igl:
Looks like they get stripped out. A stand-alone case wouldn't need it. The context of being part in a greater expression seems lost there.

vendethiel:
Any reason to close this PR? It might be the right fix, but I need to dig deeper

igl:
Woops. That was Incidental from me moving the fix to another branch on my fork. I guess i have to create a new one.

Issue #721, opened 2015-05-14
is it too much magic...in LiveScript


skandasoft:
With coffeescript being too rigid for any new request...I think it would be great to have a nice alternative like LiveScript...
I found LiveScripts offers too many things....can you limit the no of ways of doing thing(1 thing can be done in n different ways) and let the user decide his own macro(ways for using macro)..
keep it clean and simple to pick up and running..

heavyk:
forgive me for misunderstanding, but it seems your saying that coffee is too rigid, however LS is a less rigid alternative, yet still offers too many alternatives. ---- right?

so, why don't you just fork the one you like more and make it "just right"? if you look at the LS code (dunno about the coffee), it's really not that hard to add your own macros to it. I've made my own macros for LS a few times now and have been quite happy. it has been really fun to diddle around in a language and make make it my own. I removed a few things, but added others, like atumatic require module installation if not found in node_modules ... but you know those macros were important to only me cause everyone has a different way of installing their modules and stuff. fork the LiveScripts man. I'm probably going to do it again soon! there are some new things in the 1.4 that I really like. no big deal, I'll just re-add my macros. :)

vendethiel:
you might be interested in #219.

skandasoft:
@heavyk I am not here to criticize the Livescript. Just wanted to have a healthy conversation to see if it is something that people would want to have. And this conversation is not about macros..it is keep the features to minimum and let the macros handle the rest.
as for as your suggestion of forking..how about everybody forking their own way..

heavyk:
haha, I didn't think you were criticizing.

one thing that would be really useful in creating what you suggest would be some functions on the LS AST (like compile) which actually generate LS again. then, we could all utilize a common AST and heavyk-script would actually be compatible with skandasoft-script ... heavyk-script would compile to LS, then that ast would be readable by skandasoft-script. so we would essentially both be writing in our own languages, but everyone would see it as their own. the only thing necessary would be for someone to extend the ast classes and override that compile function. hmmm, interesting thought!

well, thanks for the healthy discussion, I may actually work on this. what a great idea :)
cheers

robotlolita:
I ended up never actually working on this but: https://github.com/gkz/LiveScript/issues/328#issuecomment-26763043

Issue #219, opened 2012-12-02
removing features


gkz:
So far (on master), slated for removal (and currently with compiler notices upon use):
- short function syntax - `f(x) = x` - [currently has notice]
  why: hacky implementation, duplicating functionaly
  use --> to make curried functions
- `+++` concat op [currently has notice]
  why: added ++ as concat op, which is shorter and better
- `where` statement - relatively useless - just use local var assignment or a let statement  [currently has notice]
- `undefined` alias for `void` - added for CoffeeScript compatibility, but with coffee2ls that is not an issue  [currently has notice]
- `!?` inexistence op  [currently has notice]

---

discussion:
- choose one of `**` or `^` for power op?
- remove `import` and `importAll`?
- remove `x is not y` to mean `x isnt y`, have it mean `x is !y` instead
- `yes/no/on/off` aliases - added for CoffeeScript compatibility, but now with coffee2ls, this is not an issue

More controversial:
- remove standalone use of `@`
- remove standalone or hanging use of `::`

vendethiel:
remove unjoined strings

igl:
I never use `@` and if decorators ever become a thing in JS we have a clash.
I did not have the pleasure of coding LS in a team but in coffeescribble and I found not just one bug caused by heavy usage of `?`. That disqualifies the whole thing as a safe-guard to me.

I generally prefer not using symbols if i don't have to... Not all symbols in LS have such quirky and placement specific behavior. I still prefer not to be clever just for saving a few characters.

kay-is:
I use `@` a lot. You're right with the readability and decorators aspect (even if I don't think LS has to mirror every feature of JS).
But it's a very simple substitute and for me it feels fast to skim over code and see all the `@attribute`s that are used even without special highlighting.

impinball:
@igl There's no need for decorators in LiveScript outside of changing property descriptors. You can use plain functions for nearly every use case. (I've used this in my own pet projects before)

```ls
inc = (f) -> ->
  @index += 1
  f.apply @, &

class Foo
  -> @index = 2
  foo: inc -> console.log "#{@index}"

wrap = (f) -> -> new List f.apply @, &

class List
  (@value) ->
  map: wrap (f) -> [f i for i in @value]
  filter: wrap (f) -> [i for i in @value | f i]
  each: (f) -> [f i for i in @value]; @
```

impinball:
And things that I wouldn't mind if they were removed:

[...]

PR #803, opened 2015-12-30
Start working on type ascriptions

vendethiel:
I would like everyone's input on this: the syntax, what we're compiling to, etc.
The idea is to compile down to TS or flow-compatible stuff

### DIFF
--- a/src/ast.ls
+++ b/src/ast.ls
@@ -535,6 +535,8 @@ class exports.Var extends Atom
 
     maybe-key: -> Key(@value) <<< {@line}
 
+    set-type: (@type) -> this
+
     var-name: ::show
 
     compile: (o) -> sn(this, if @temp then o.scope.free @value else @value)
@@ -1915,7 +1917,9 @@ class exports.Fun extends Node
                     vr = v
                 else if df
                     assigns.push Assign vr, p.second, \=, p.op, true
-                names.push (scope.add vr.value, \arg, p), ', '
+                names.push (scope.add vr.value, \arg, p)
+                names.push " : #that" if vr.type?
+                names.push ', '
         if rest
             while splace-- then rest.unshift Arr!
             assigns.push Assign Arr(rest), Literal \arguments
--- a/src/grammar.ls
+++ b/src/grammar.ls
@@ -157,6 +157,7 @@ bnf =
     ArgList:
         o ''                                                -> []
         o 'Arg'                                             -> [$1]
+        o 'Arg ASCR ID'                                     -> [$1.set-type $3]
         o 'ArgList , Arg'                                   -> $1 ++ $3
         o 'ArgList OptComma NEWLINE Arg'                    -> $1 ++ $4
         o 'ArgList OptComma INDENT ArgList OptComma DEDENT' ditto
--- a/src/lexer.ls
+++ b/src/lexer.ls
@@ -652,6 +652,8 @@ exports <<<
             @adi!
             val = 'prototype'
             tag = 'ID'
+        case '@:'
+            tag = 'ASCR'
         case '=>'
             @unline!
             @fset 'for' false
@@ -1343,6 +1345,7 @@ SYMBOL = //
 | \([^\n\S]*\)                  # call
 | [!=]==?                       # strict equality, deep equals
 | !?\~=                         # fuzzy equality
+| @:                            # type ascription
 | @@?                           # this / constructor
 | <\[(?:[\s\S]*?\]>)?           # words
 | <<<<?                         # import


heavyk:
could you give an example? I'm not sure I understand

robotlolita:
So, basically:

```ls
(a @: Type, b @: Type) -> ...
```

Gets compiled to:

```js
(function(a : Type, b : Type){ ... })
```

Is the idea to support things like TypeScript or Flow?

vendethiel:
Yes, indeed.

robotlolita:
Ah. I'm not particularly fond of the `@:` symbol, but I also can't think of any other that is available and denotes a typing relationship well :(

vendethiel:
Right? `:` and `::` both doesn't work, so... I went for the one I could think of.

heavyk:
oh, I see now. yea, I like the idea of compiling to TS/flow.

could this work?
```ls
fn = (:a Type, :b Number) ->
```

vendethiel:
It'd make the lexer a bit more complex, which I'm not too fond of. But I'm open to opinions.

robotlolita:
An alternative might be using the same symbol Clojure uses to declare types. It's reusing a symbol, but I don't think that's an issue for argument names:

```ls
foo = (a ^Type, b ^Another-Type) ^ReturnType -> ...
foo = (a ^Type = ^clone) -> ...
```

vendethiel:
Yeah, but it's currently parsed as an infix MATH (power) :/

vendethiel:
At least, the fact our unary clone got changed to `^^` helps (because we allow unary ops there)

waynedpj:
> Right? : and :: both doesn't work, so... I went for the one I could think of.

is there any interest in making some breaking changes to free up operators like `:` and `::` that many already recognize as defining types, particularly `:` used in TypeScript+Flow?  IMHO the type ascriptions would take precedence over most other uses of operators.

also, even the Livescript homepage seemingly uses `::` for type ascriptions (in the 'Programmatic API' section):
`LiveScript.compile(code :: String, options :: Object?) -> String`

finally, there already seems to be some discussion on removing duplicated/etc. functionality:
https://github.com/gkz/LiveScript/issues/219
https://github.com/gkz/LiveScript/issues/721

peace

vendethiel:
We can change the docs, that's very minor. It currently just mimics Haskell's syntax :-).

You're free to ping those issues to try and get some tractions ;-).

gkz:
(x :: Number) ->

Seems to be a syntax error - maybe we could just support the syntax for function definitions?

vendethiel:
Problem is just the lexer's complexity. We need to resort to per-nest flag (or something) to parse correctly `(a :: Foo = -> ::) ->`, say

vendethiel:
Urgh, https://github.com/gkz/LiveScript/blob/master/test/cs-classes.ls#L289 doesn't parse correctly. Obviously. I guess I need to change the regexp to `@:(?!:)`...

heavyk:
ok, if you don't like the atom syntax, what about c++ syntax?

```ls
lala = (a<Type>, b<Number> = 3) -> ...
```

rightfold:
`a isa number`

igl:
Arrow functions are fine and dandy but what aboot  braceless function declarations? 
```
function add : number a : number, b : number => a + b
function add @ number a @ number, b@number => a + b
function add @: number a @:number, b @:number => a + b
function add~number a~number, b~number => a + b
function add^number a^number, b^number => a + b
function add<number> a<number>, b<number> => a + b
```

C++ style looks best to me.

What do you guys think about type contracts like in contracts.coffee or elm?

vendethiel:
I really don't like parenless function definitions, but anyway, all the signatures will look the same.

unclechu:
I vote for `::` it's easy to catch by eyes for me and I think this:
```text
fn = (a :: Type, b :: Number) ->
```
is more obvious than that:
```text
fn = (:a Type, :b Number) ->
```

coachshea:
agreed

igl:
Unfortunately `::` compiles to `prototype` in LS.
There is also a es7 proposal using `::` it's only a proposal though.

AriaMinaei:
How about declaring the function type in a separate line?

```livescript
add :: (Number, Number) -> Number
add = (a, b) ->
  a + b
```

I think the other syntax (defining the type in the same line as declaring the function) creates too much visual clutter, and it just doesn't go with LiveScript's terse syntax, but that might be a matter of taste.

unclechu:
@AriaMinaei may be as alternative, but if you have a lot of arguments, inline declaration will be easier to read.
```livescript
add :: (
  number
  number
  string
  boolean
  Object
  Date
  string[]
) -> number
add = (
  foo
  bar
  test
  test2
  abc
  def
  zyx
) ->
  a + b
```
In this case I can't say immediately what type does have `test2`.
And in case with inline type declaration we keep it short and readable:
```livescript
add = (
  :foo   number
  :bar   number
  :test  string
  :test2 boolean
  :abc   Object
  :def   Date
  :zyx   string[]
): number ->
  a + b

**Comment by AriaMinaei** [comment]
@unclechu Correct, but how many functions *would* have that many arguments? A function invocation with more than five or six arguments would be a bit unreadable, unless the arguments are named (through a config object).

Also, writing type annotations in the same line as function declaration makes even simple functions like [this](https://github.com/AriaMinaei/shapely/blob/master/src/shapely/validators/ExactValueValidator.js#L18) overly verbose. I'm not even sure if that *is* a matter of taste.

**Comment by focused** [comment]
I love this one style of type declaration:
```
add :: (Number, Number) -> Number
add = (a, b) ->
    a + b
```
@unclechu I think we should not pass more than 4 or 5 arguments, in this case we should use data structures (Array, Object). Look:
```
count-leaves :: (Tree, BranchSide) -> Number
count-leaves = (t, s) ->
  ...
```

isiahmeadows:
@vendethiel Status?

vendethiel:
@isiahmeadows still undecided on which token. Overloading `::` seems risky (and hard to parse in cases like `(a = -> x::) ->`, say).

thelambdaparty:
Maybe [JavaScript types in TypeScript](https://github.com/Microsoft/TypeScript/issues/4790) ?
Uses JSDoc for type information.

isiahmeadows:
@thelambdaparty I really dislike JSDoc type annotations. They are almost a joke in their verbosity.

rightfold:
> They are almost a joke

As is TypeScript.

isiahmeadows:
@rightfold To be honest, *Java's* type system are better than JSDoc annotations in my opinion. And I really don't like Java. The only thing they're worth using for are for external API documentation, and even that's disputable.

I wasn't comparing them to powerful type systems like Haskell's.

rightfold:
Who said anything about Java?

isiahmeadows:
I was drawing a comparison to clarify what I meant. Basically, I feel JSDoc
types would suck more than the type system of a really bad language.
Especially for a concise language.

kress95:
So many options...

    (a & Type, b & Type) -> ...
    (a  @: Type, b @: Type) -> ...
    (a ::{Type}, b ::{Type}) -> ...
    (a ::<Type>, b ::<Type>) -> ...
    (a &<Type>, b &<Type>) -> ...
    (a <:Type:>, b <:Type:>) -> ...
    (a {:Type:}, b {:Type:}) -> ...
    (a {.Type.}, b {.Type.}) -> ...

(I really like the syntax of the last two, reminds me of nim's pragmas)

isiahmeadows:
I'm okay with any of those except the ones with ampersands (I would
consistently misread them as type union, which TypeScript isn't the only
one using that syntax).

kay-is:
Do they need to be non-valid js identifiers (like : or ^ etc.)? 

If they can be anything, how about `§` ?

`§` is a bit associated with laws (which types essentially are)

    my-func = (a §Number, b §String) §Boolean -> 
    my-func = (a §Number, b §String) -> §Boolean 

    my-func = §Boolean (§Number a, §String b)->

isiahmeadows:
I would prefer ASCII symbols to denote types, BTW. And preferably, yes,
they should. I already run into issues occasionally with `is` and `_`.

heavyk:
I have the § key on my keyboard and I would have suggested it also but the us kbd does not have the key. I always thought it was ascii, but upon looking - alas it is not - I guess it was one of the iso standards.

I'm also ok with whatever cept for ampersands. I suppose the clearest to my eyes is the `a @: Type` one

focused:
I have no such key, and many of us

heavyk:
sorry, that wasn't the point.

> I suppose the clearest to my eyes is the `a @: Type` one

rhendric:
I know this is a closed PR, but since @dk00 is still linking to it, it seems useful as a repository for bikeshed suggestions. Here's mine:

`(a (Number), b (String)) (Boolean) -> ...`
`x (foo: String || Number) = get-foo!`

Advantage of this syntax over most of the above: it already parses into an AST, and definitely won't conflict with any valid LS since you can't assign to a function call.

vendethiel:
I'm not sure what that 2nd line describes?

rhendric:
Ascriptions for variable declaration.

vendethiel:
is `x` or `foo` the variable?

rhendric:
`x` is a variable, with a type that asserts it has a property named `foo` with value that is a `String` or a `Number`. As elsewhere in LS, `foo: String || Number` is shorthand for `{ foo: String || Number }`.

vendethiel:
Isn't this ambiguous with `x((foo: String || Number) = get-foo!)` (declare a variable inside a funcall)?

rhendric:
Gah, yes, you're right! Righto, that's what I get for tossing off half-baked thoughts in passing.

vendethiel:
It might be better, at least at first, to only allow type ascription when using declarators (`var`/`const`) (maybe in `let`s as well).

ozra:
Another soft ball in the air: `foo = (x ^Number, y ^String) ^String -> "#{x}, #{y}"`.

Too bad @kay-is suggestion of `§` isn't us-keymap and us-charset friendly, I considered that one in my pet lang before too :-/

The Haskellish way posed earlier in the issue is a good bet too, and when non-clashing symbology can be figured out - _both_ the inline- and pre- typedecl would be awesome imo.

vendethiel:
Well, there's always the solution of writing the type signature before the object itself, and not inline. That'd remove a good chunk of ambiguities.

danielo515:
What is the status of this ?

rhendric:
Exactly what it looks like. There's the occasional idea being tossed around in #999, if you feel like helping out.

danielo515:
I'll take a look at that issue.
Thanks

PR #806, opened 2016-01-07
fixes #634 : Partial application should be bound (like bound functions)

### LINKED ISSUE #634: Partial application should be bound (like bound functions)
It's really confusing to have code like this constantly throw TypeErrors:

```ls
class Walker
    (@visitors) ->

    walkList: (list) !->
        list
        |> filter (.type of @visitors)
        |> each @walk

    # more methods
```

The problem is this partial function:

```ls
(.type of @visitors)
```

It looks like it should be bound to the current instance, equivalent to this code:

```ls
(node) ~> node.type of @visitors  # bound
```

It's really equivalent to this, which was a rather unexpected gotcha:

```ls
(node) -> node.type of @visitors  # not bound
```

This gotcha is a little misleading. It feels like a small fix, but I don't know a lot about the internal compiler structure.

summivox:
#634

### DIFF
--- a/src/ast.ls
+++ b/src/ast.ls
@@ -1467,9 +1467,9 @@ class exports.Binary extends Node
             y = Var \y$
             sn(this, (Fun [x, y], Block((Binary @op, x, y).invert-check this), false, true).compile o)
         case @first?
-            sn(this, "(", ((Fun [vit], Block((Binary @op, @first, vit) .invert-check this)).compile o), ")")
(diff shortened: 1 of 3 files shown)


vendethiel:
Can you please add some tests?

summivox:
@vendethiel done

vendethiel:
Still waiting for travis :\

PR #808, opened 2016-01-10
Fixes #739: can not compile empty list of words

### LINKED ISSUE #739: can not compile empty list of words
If I have an empty list of words like this

```
<[
]>
```

the lexer throws `TypeError: Cannot read property '0' of undefined`. This only happens if the only character in brackets is the newline. Multiple newlines and whitespace are ok.

summivox:
reason: unchecked array bound

### DIFF
--- a/src/lexer.ls
+++ b/src/lexer.ls
@@ -1281,7 +1281,7 @@ character = if not JSON? then uxxxx else ->
                 tokens[i].0 = '+-'
             continue
         default continue
-        if token.spaced and tokens[i + 1].0 in ARG
+        if token.spaced and tokens[i + 1]?.0 in ARG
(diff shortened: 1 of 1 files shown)


vendethiel:
Thanks for the PR, but it seems to me the `case WORDS` earlier shouldv've caught this, no?

summivox:
IIRC not when `WORDS` is the final token.
These are strange cases anyway... for example:
```livescript
<[
]>
alert!
```
I just discovered this -- it fails at AST step.
---
Sure, in theory you should have that compile to `[]` but why bother?

vendethiel:
Because people are going to expect `x = <[]>` to work.

Issue #562, opened 2014-09-09
Lsc hangs when executing shelljs.exec from the shelljs library [lsc issue only, works when precompiled]


gkovacs:
Using LiveScript 1.2.0 on Mac OS X 10.9.4, nodejs v0.10.21

Shelljs: https://github.com/arturadib/shelljs (I'm not affiliated with them, just trying to use it). Also reported to them at https://github.com/arturadib/shelljs/issues/160 in case this is some issue with their code

A test case is below, executing it either in the interactive LiveScript interpreter or by saving it to test.ls and running with lsc test.ls causes it to hang after outputting "foo" and "bar" (so it's hanging on the shell.exec call).

console.log 'foo'
shell = require('shelljs')
console.log 'bar'
console.log shell.exec('ls')
console.log 'baz'

To test, save to test.ls and run with lsc test.ls.

Interestingly the code above works fine when pre-compiling the LiveScript to Javascript and executing with node, so I think this issue is exclusive to lsc: "lsc -c test.ls && node test.js" runs correctly.

gkovacs:
So I've determined why this issue occurs. shelljs uses process.execPath to determine where the node interpreter is, as you can see in https://github.com/arturadib/shelljs/blob/master/src/exec.js#L47

Both node and coffee return the path to the node interpreter (in my case, /usr/local/bin/node) as process.execPath. However, lsc returns /usr/local/bin/lsc as process.execPath. Is there a good reason why lsc should behave differently from coffee for giving the lsc path (as opposed to the node path) for process.execPath? It certainly does appear to be intentional, as I see in https://github.com/gkz/LiveScript/blob/master/src/command.ls#L45

```
process.exec-path = process.argv.0 = process.argv.1
```

Anyhow, substituting that instance of process.execPath with /usr/local/bin/node in the shelljs codebase fixes the isssue. So does removing that line in command.ls

gkovacs:
Also, if for whatever reason it is deemed necessary to have process.execPath point to lsc instead of node, how can one get the path to the node interpreter when executing a script from lsc?

vendethiel:
interesting...

PR #822, opened 2016-01-24
make process.execPath point to node instead of lsc

gkovacs:
When running a script in lsc, process.execPath points to lsc instead of node. This leads to issues with libraries such as shelljs: https://github.com/gkz/LiveScript/issues/562 and https://github.com/shelljs/shelljs/issues/160


vendethiel:
@gkz? This seems like a more serious issue than I first thought -- sorry @gkovacs

heavyk:
+1 it's really annoying, yeh

vendethiel:
Thanks!

PR #823, opened 2016-01-25
fix: `yield` expression should be callable

summivox:
Since `yield` expressions evaluate to argument passed to `next`, if a function is passed it should be callable.

Example:

```livescript
f = ->* (yield)!
g = -> console.log 'hit'
f!
  ..next!
  ..next g
```

should compile to:

```javascript
var f, g, x$;
f = function*(){
  return (yield)();
};
g = function(){
  return console.log('hit');
};
x$ = f();
x$.next();
x$.next(g);
```

### DIFF
--- a/src/ast.ls
+++ b/src/ast.ls
@@ -1124,6 +1124,8 @@ class exports.Yield extends Node
 
     show: -> if @op is 'yieldfrom' then 'from' else ''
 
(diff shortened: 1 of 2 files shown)


vendethiel:
Please add some tests :).

summivox:
@vendethiel done

vendethiel:
Thanks!

PR #831, opened 2016-01-31
Support unary operator spread over object literals

rhendric:
<table>
<thead>
<th>Livescript</th>
<th>Compiled JavaScript</th>
</thead>
<tbody>
<tr>
<td>
<pre>x = delete a{b, c: d}</pre>
</td>
<td>
<pre>
var x, ref$;
x = {
  b: (ref$ = a.b, delete a.b, ref$),
  c: (ref$ = a.d, delete a.d, ref$)
};
</pre>
</td>
</tr>
<tr>
<td>
<pre>x = do loaders{plugins, preferences: {server, client}}</pre>
</td>
<td>
<pre>
var x;
x = {
  plugins: loaders.plugins(),
  preferences: {
    server: loaders.server(),
    client: loaders.client()
  }
};
</pre>
</td>
</tr>
</tbody>
</table>

Works almost exactly like unary spread over list literals/slices, but with that tasty object flavor.

I was originally going to support all unary operators, but then had a crisis of conscience when I looked at `^^{foo, bar}` and asked whether that ought to be equivalent to `x = {foo, bar}; ^^x` or `{^^foo, ^^bar}`, or whether someone would freak out if `typeof {} != typeof {foo:1}` (although both of these issues exist with list spreads right now, so...). The somewhat arbitrary list of operators I decided to support is `~` `!`/`not` `new` `do` `delete`/`delete!` `++` `--`, but I'm not terribly attached to most of those decisions. The two examples demonstrated above are the ones I want to use.

Feedback please!

### DIFF
--- a/lib/ast.js
+++ b/lib/ast.js
@@ -1932,13 +1932,22 @@ exports.Unary = Unary = (function(superclass){
     }
   };
   prototype.compileSpread = function(o){
-    var it, ops, them, i$, len$, i, node, sp, j$, op, lat, ref$;
+    var it, ops, isSpread, isObj, them, i$, len$, i, node, sp, j$, op, lat, ref$;
     it = this.it;
     ops = [this];
     for (; it instanceof constructor; it = it.it) {
       ops.push(it);
     }
-    if (!((it = it.expandSlice(o).unwrap()) instanceof Arr && (them = it.items).length)) {
+    isSpread = (it = it.expandSlice(0).unwrap()) instanceof Arr || (isObj = it instanceof Obj && !function(){
+      var i$, x$, ref$, len$, ref1$;
+      for (i$ = 0, len$ = (ref$ = ops).length; i$ < len$; ++i$) {
+        x$ = ref$[i$];
+        if ((ref1$ = x$.op) !== '~' && ref1$ !== '!' && ref1$ !== 'new' && ref1$ !== 'do' && ref1$ !== 'delete' && ref1$ !== 'jsdelete' && ref1$ !== '++' && ref1$ !== '--') {
+          return true;
+        }
+      }
+    }());
+    if (!(isSpread && (them = it.items).length)) {
       return '';
     }
     for (i$ = 0, len$ = them.length; i$ < len$; ++i$) {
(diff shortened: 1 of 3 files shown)


vendethiel:
At that point, I think I'd prefer if there were an operator to trigger it.

rhendric:
For object literals/slices only (inconsistent) or for list literals/slices also (breaking)?

FWIW, `new {a}`, `do {a}`, `delete {a}`, and the four inc/dec pre/post flavors of `{a}++` (as well as more complex slicey versions of each) are all compiler errors right now; restricting to just those would result in minimal ambiguity.

Do we really want to live in a world where `a{b, c} += 1` works, `a[1, 2]++` works, but `a{b, c}++` doesn't work, or needs an extra operator to trigger it?

vendethiel:
The breaking one.

rhendric:
Hmm. How about if we marked the literal/slice to spread over with a splat?

`new ...[circle, square]` => `[new circle, new square]`
`++...{level, max: max-level}` => `{level: ++level, max: ++max-level}`
`...player{strength, hp}--` => `{strength: player.strength--, hp: player.hp--}`

Extend to assignment operators?
`...player{strength, hp} += 5` => `{strength: player.strength += 5, hp: player.hp += 5}`

vendethiel:
I'd say it looks consistent with what spread does -- but not sure how to parse it OTTOMH

rhendric:
The parsing doesn't appear to be a problem; I bashed out a quick proof-of-concept. Is this proposal baked enough for me to harden the implementation into a new PR? Should I leave support for existing spread syntax intact, add a deprecation warning, or just remove it?

vendethiel:
Well, that's where I'd have liked @gkz's input.

I guess we can start to consider master to be LS2's head (maybe I'll create an actual LS2 dev branch later), so what I'd like to do before merging is just... Ask people!

If you could please post a message with a note of such a change in the LS2 threads. I hope we won't find any place where that's ambiguous (shouldn't be)

isiahmeadows:
@vendethiel [I see a branch prime for a push.](https://github.com/gkz/LiveScript/tree/ls2) The readme can be updated and posted on the new, replaced README. It only has one commit.

vendethiel:
No, not that branch. If anything, it should be deleted and recreated from current master.

vendethiel:
@rhendric I'm gonna go ahead and say "let's go for the explicit spread syntax". So long `foo ...+...a<[b c d]>` works

PR #832, opened 2016-01-31
Consistently check whether the map option as been supplied to compile

msteen:
Compile returns an object instead of a string if supplied the map option, so this has to be checked before using the returned value.


vendethiel:
We should probably add CLI tests...

Thanks!

PR #833, opened 2016-01-31
Made the checks for compiling with the map option more consistent

msteen:


src/index.ls
```diff
@@ -44,7 +44,7 @@ exports <<<
                   output = new SourceNode null, null, null, [
                       "// Generated by LiveScript #{exports.VERSION}\n", output
                   ]
-              if options.map and options.map isnt 'none'
+              if options.map? and options.map isnt 'none'
```
vendethiel: It should be a string if it's present, not `false`, right?

msteen: Yes, that is right, `options.map` should be a string, 'none', 'embedded', etc. If you now passed `options.map = 5`, it would be as if it were `embedded`, my change did not effect this behavior. I only changed it because I thought it to be better LiveScript code to use `value?` for checking the existence of some value, because it makes the intent clear.

If the code is only for control flow, than how it currently is should suffice. Then you assume the value to be valid. If you also want to do validation, the check would become:`if options.map in <[ linked linked-src embedded debug ]>`. But than you will have to change a lot more code, because validation of options is currently not being done except for the CLI.

rhendric:
I don't think the refinement in this PR is necessary. If we were to support `map: false` in `options`, I would argue that its behavior ought to be the same as `map: 'none'`.

PR #835, opened 2016-01-31
CLI: When on Windows, print Windows path separator `\` instead

summivox:
reason: consistency; passing test

### DIFF
--- a/src/command.ls
+++ b/src/command.ls
@@ -203,6 +203,7 @@ switch
   dir = path.dirname source
   if o.output
       dir = path.join that, dir.slice if base is '.' then 0 else base.length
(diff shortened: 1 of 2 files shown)


vendethiel:
Breaks the build

summivox:
Ah damn. It seems that I have used APIs available only in the newer Node versions, namely `path.posix.join`.

Question: Would you fix the test to include windows separator `\`, or make CLI output `/` ?

summivox:
@vendethiel : phew, now it passes even on node `v0.10.x`. Shouldn't have taken new Node APIs for granted ;)

vendethiel:
Thanks!

Issue #811, opened 2016-01-12
So-called "next big version"


vendethiel:
LiveScript "2.0"
============

"It's a-boot time"

So, this got talked about quite about before, but I'm not sure that's actually what's interesting to us now.

However, there's definitely a need to cleanup some things in LS, and especially with ES6/7-related.

Are people okay with breaking changes now? If they're not, we can discuss what we want to change, but in the meantime:

- [ ] Implement type ascriptions (PR starting #803)
- ES6 support:
	- modules. @gkz proposed `import!` and `export!` for the ÈS6 modules (we need to decide between `<<<{<,}` and `import {all,}` anyway)
	- classes: **I don't think we want them**. LiveScript code tends to use invokable class bodies
	- async/await can be delayed a bit more, I think. It's ES7, and it's not quite there yet... We'd also need to decide on a syntax (eww `->**`)
	- `const` needs to be compiled to ES6's `const`
	- I'm not sure we need to change `let`'s meaning, I don't think it adds much.
- [ ] Remove `for ever`
- [ ] Parse the star as a real operator (not just void=)
- [ ] Remove `implements` (unless people actually use it)
- [ ] Remove the match-introduced operators `>==` etc
- [ ] Make `match` an actual pattern matching operation (semantics TBD, that can come later)
- [ ] Remove operator overloading (...or make the compiler aware of types...)
- [ ] Trim down synonmys. Includes `otherwise`/`default`, `case`/`when`, `=>`/`then` (I guess this one is the one people like), etc

This will be controversial (at best), but I think it's worth discussing:
I believe we should remove implicit variable declaration. We have `var`, and `:=` when we need to change a value.
I think the scoping rules would seem much saner if we had to explicitly declare those. But I know this isn't going to fly for a lot of people, so...

There's another controversial topic: `..`, `it` and `that`. I like what we currently have, but I'm too biased to know if that bugs people or not.

gkz:
Good idea, let's at least get everything down on a list.

igl:
Types are neat.. but does that mean there will be `--compile-target es6/ts/flow`?
I would like to see object-spread/rest on this list. `<<<`/`import` would be obsolete.

++ es6 modules, const, match, it
-- implicit variable declaration, that and everything else!

rightfold:
If you have to pick either from TS or Flow, and are not able to support both, definitely pick Flow. It's sound. Supporting both would be nice though.

darrrk:
[...]

Issue #754, opened 2015-07-11
Using backcalls with Promises


mrmayfield:
I have some code that runs a Promise on repeat.

'''
putOnRotation = (callback, delay) ->
  return callback delay
  .then (data) ->
    console.log(data)
    return Promise.delay(delay).then ->
      return putOnRotation callback, delay
  .catch (error) ->
    console.log(error)
    return Promise.delay(5000).then ->
      return putOnRotation callback, delay
'''

However, I don't like this pyramid pattern, and would prefer to use backcalls.

I've ran into a litany of issues getting them set up, any ideas on how I could remove the indentation of this function using backcalls? Thanks.

robotlolita:
```ls
put-on-rotation = (f, delay) ->
  do
    data <- f delay .then
    console.log data
    _ <- Promise.delay delay .then
    put-on-rotation f, delay
  .catch (err) ->
    console.log err
    _ <- Promise.delay 5000 .then
    put-on-rotation f, delay    
```

mrmayfield:
Thanks a lot! But what are you using for syntax highlighting? I am using this Sublime Text bundle: https://github.com/paulmillr/LiveScript.tmbundle but the highlighting looks pretty horrific haha. Missing most of LS features and promises are not well understood either, any better suggestions?

robotlolita:
I don't use LS anymore (used to use the Emacs mode back then). Don't know any tool with good support for LS, maybe @vendethiel knows?

vendethiel:
I havn't used LS for a while (I've thankfully been able to stay away for JS), but I was using the tmbundle as well. I know the vim mode works rather well.

gabeio:
I currently Atom.io with @blvz's livescript highlighter `language-livescript-edge` idk if it has all the features highlighted but it's sufficient for me :) :+1:

iyel:
@robotlolita why don't u use LS anymore?

blvz:
@gabeio I don't know either. Hahaha.

@mrmayfield Take a look at this: https://github.com/paulmillr/LiveScript.tmbundle/pull/39

robotlolita:
@iyel I'm using Sweet.js for writing JS (because macros), and otherwise concentrating the efforts on my own PLs.

heavyk:
@robotlolita what sorts of macros? wouldn't it be possible to do the same in LS? they don't look *that* difficult to emulate (I think)

vendethiel:
[...]

Issue #752, opened 2015-07-09
Promise Generators


kkirby:
Hello.

I have updated LiveScript in a fork to support promise generators. The syntax looks like this:

	wait = (time) -> new Promise (fulfill) -> setTimeout fulfill, time

	my-promise = ->**
		console.log \helloWorld
		yield wait 1000
		console.log \goodbyeWorld

	function my-promise **()
		console.log \helloWorld
		yield wait 1000
		console.log \goodbyeWorld

The compiled code renders as such:

    // Generated by LiveScript 1.4.0
    (function(){
      var wait, myPromise;
      wait = function(time){
        return new Promise(function(fulfill){
          return setTimeout(fulfill, time);
        });
      };
      myPromise = generatorToPromise$(function*(){
        console.log('helloWorld');
        (yield wait(1000));
        return console.log('goodbyeWorld');
      });
      function tryCatch$(func,context,args){
          try {
              return {
                  failed: false,
                  value: func.apply(context,args)
              };
          }
          catch(exception){
              return {
                  failed: true,
                  value: exception
              }
          }
      }
      function generatorToPromise$(func,argumentLength){
          function processGenerator(generator){
              return new Promise(
                  function(fulfill,reject){
                      function next(passIn,handler){
                          handler = handler || 'next';
                          var tryCatchResult = tryCatch$(generator[handler],generator,[passIn]);
                          if(tryCatchResult.failed === true){
                              return reject(tryCatchResult.value);
                          }
                          var result = tryCatchResult.value;
                          if(result.done){
                              return fulfill(result.value);
                          }
                          else if(result.value instanceof Promise){
                              result.value.then(
                                  next,
                                  function(promiseError){
                                      next(promiseError,'throw');
                                  }
                              );
                          }
                          else {
                              next(result.value);
                          }
                      }
                      next()
                  }
              );
          }
          if(typeof func != 'function' && typeof func.next == 'function'){
              return processGenerator(func)
          }
          else {
              if(typeof argumentLength == 'number'){
                  var args = [];
                  for(var i = 0; i < argumentLength; i++){
                      args.push('arg'+i);
                  }
                  args = args.join(',');
                  return new Function('processGenerator,func','return function(' + args + '){ return processGenerator(func.call(this,' + args + ')); };')(processGenerator,func);
              }
              else {
                  return function(){
                      return processGenerator(func.apply(this,arguments));
                  }
              }
          }
      }
    }).call(this);

[...]

Issue #764, opened 2015-08-08
Questionable semantics: list comprehension in `yield` expression


summivox:
Closest issue I can find is  #667 

```livescript
->* yield [x for x in arr]
```
compiles to:
```javascript
(function*(){
  var x;
  return (yield (yield* (function*(){
    var i$, ref$, len$, results$ = [];
    for (i$ = 0, len$ = (ref$ = arr).length; i$ < len$; ++i$) {
      x = ref$[i$];
      results$.push(x);
    }
    return results$;
  }())));
});
```

```livescript
->* yield alert [x for x in arr]
```
compiles to:
```
(function*(){
  var x;
  return (yield alert((yield* (function*(){
    var i$, ref$, len$, results$ = [];
    for (i$ = 0, len$ = (ref$ = arr).length; i$ < len$; ++i$) {
      x = ref$[i$];
      results$.push(x);
    }
    return results$;
  }()))));
});
```

While in all cases the behavior is as expected, the compilation is surprising and *undocumented*. Clearly the `yield*` is extraneous in both cases. I suspect there must have been use cases where the `yield*` makes sense, but could we at least document this, and clarify when this rule is applicable?

gkovacs:
This bug appears with switch statements as well:

```
(name) ->*
  x = switch name
  | 'a' => 0
  | 'b' => 1
  return x
```

compiles to

```
// Generated by LiveScript 1.5.0
(function(){
  (function*(name){
    var x;
    x = (yield* (function*(){
      switch (name) {
      case 'a':
        return 0;
      case 'b':
        return 1;
      }
    }()));
    return x;
  });
}).call(this);
```

PR #836, opened 2016-02-01
ES7 `async`/`await` implementation with concise syntax

summivox:
[ES7 async/await](http://wiki.ecmascript.org/doku.php?id=strawman:async_functions) is a syntax sugar for promise+generator async control pattern. Although this feature is still a draft in even ES7, it has gained considerable traction, and is completely usable right now through [a babel plugin](http://masnun.com/2015/11/11/using-es7-asyncawait-today-with-babel.html).

This patch provides a concrete example on which we can discuss syntax extensions to LiveScript that compiles to ES7 async/await.
# `async function`
1. Append an additional `>` to all arrows: `->>`, `~>>`, `-->>`, `~~>>`
   Rationale for the syntax: terseness
2. ~~Prepend keyword `async` to arglist of arrow functions: `async (arglist) -> body`
   Rationale for the syntax: agrees with [strawman](http://wiki.ecmascript.org/doku.php?id=strawman:async_functions) proposal on async "ES6 rocket" functions~~ (redundant)
3. Prepend keyword `async` to named function definitions: `async function name(arglist)`
   Rationale for the syntax: obvious
# `await` expression
1. `await` compiles to `await` (in the same way as how `yield` is compiled now)
2. ~~`await all` compiles to `await*` (in the same way as how `yield from` is compiled now)
   Rationale:  `await*` is proposed to have the same semantics as `Promise.all` (see [strawman](http://wiki.ecmascript.org/doku.php?id=strawman:async_functions))~~ (no longer in standard)
# Example

``` livescript
af1 = (x) ->> await x
af2 = async (a, b, c) ~> await @x
async function af3(x)
  a = await af1 x
```

compiles to

``` javascript
var af1, af2, this$ = this;
af1 = async function(x){
  return (await x);
};
af2 = async function(a, b, c){
  return (await this$.x);
};
async function af3(x){
  var a;
  return a = (await af1(x));
}
```
# References
- [Official Strawman Proposal](http://wiki.ecmascript.org/doku.php?id=strawman:async_functions)
- [Example of async/await usage](http://code.tutsplus.com/tutorials/a-primer-on-es7-async-functions--cms-22367)
- [Babel plugins for transpiling ES7 async/await to ES5](http://masnun.com/2015/11/11/using-es7-asyncawait-today-with-babel.html)
- #752 #754 : prior art
- #811 : LiveScript "2.0" checklist

### DIFF
--- a/src/ast.ls
+++ b/src/ast.ls
@@ -1151,21 +1151,23 @@ class exports.Yield extends Node
 
     children: <[ it ]>
 
-    show: -> if @op is 'yieldfrom' then 'from' else ''
+    show: -> switch @op
+    | 'yield' => ''
+    | 'yieldfrom' => 'from'
+    | 'await' => 'await'
+    | 'awaitall' => 'await all'
+    | _ => ''
 
     ::delegate <[ isCallable ]> -> yes
 
     compile-node: (o) ->
-        code = []
-
-        if @op is \yieldfrom
-            code.push 'yield*'
-        else
-            code.push 'yield'
-
-        if @it
-            code.push " #{@it.compile o, LEVEL_OP + PREC.unary}"
-
+        code = [(switch @op
+        | 'yield' => 'yield'
+        | 'yieldfrom' => 'yield*'
+        | 'await' => 'await'
+        | 'awaitall' => 'await*'
+        )]
+        if @it then code.push " #{@it.compile o, LEVEL_OP + PREC.unary}"
         sn(this, "(", ...code, ")")
 
 #### Unary operators
@@ -1881,7 +1883,7 @@ class exports.Existence extends Node implements Negatable
 #### Fun
 # A function definition. This is the only node that creates a `new Scope`.
 class exports.Fun extends Node
-    (@params or [], @body or Block!, @bound and \this$, @curried or false, @hushed = false, @generator = false) ~>
+    (@params or [], @body or Block!, @bound and \this$, @curried or false, @hushed = false, @generator = false, @async = false) ~>
 
     children: <[ params body ]>
 
@@ -1913,7 +1915,11 @@ class exports.Fun extends Node
         o.indent += TAB
         {body, name, tab} = this
         code = [\function]
-        if @generator
+        if @async
+            @ctor and @carp "a constructor can't be async"
+            o.in-generator = true
+            code.unshift 'async '
+        else if @generator
             @ctor and @carp "a constructor can't be a generator"
             o.in-generator = true
             code.push \*
--- a/src/grammar.ls
+++ b/src/grammar.ls
@@ -277,11 +277,15 @@ bnf =
 
         # The function literal can be either anonymous with `->`,
         o 'PARAM( ArgList OptComma )PARAM -> Block'
-        , -> Fun $2, $6, /~/.test($5), /--|~~/.test($5), /!/.test($5), /\*/.test($5)
+        , -> Fun $2, $6, /~/.test($5), /--|~~/.test($5), /!/.test($5), /\*/.test($5), />>/.test($5)
+        o 'ASYNC PARAM( ArgList OptComma )PARAM -> Block'
+        , -> Fun $3, $7, /~/.test($6), /--|~~/.test($6), /!/.test($6), false, true
         # or named with `function`.
         o 'FUNCTION CALL( ArgList OptComma )CALL Block' -> (Fun $3, $6).named $1
         o 'GENERATOR CALL( ArgList OptComma )CALL Block'
-        , -> (Fun $3, $6, false, false, false, true).named $1
+        , -> (Fun $3, $6, false, false, false, true, false).named $1
+        o 'ASYNC FUNCTION CALL( ArgList OptComma )CALL Block'
+        , -> (Fun $4, $7, false, false, false, false, true).named $2
 
         # The full complement of `if` and `unless` expressions
         o 'IF Expression Block Else'      -> L 1 2 If $2, $3, $1 is 'unless' .add-else $4
--- a/src/grammar.ls
+++ b/src/grammar.ls
@@ -278,8 +278,6 @@ bnf =
         # The function literal can be either anonymous with `->`,
         o 'PARAM( ArgList OptComma )PARAM -> Block'
         , -> Fun $2, $6, /~/.test($5), /--|~~/.test($5), /!/.test($5), /\*/.test($5), />>/.test($5)
-        o 'ASYNC PARAM( ArgList OptComma )PARAM -> Block'
-        , -> Fun $3, $7, /~/.test($6), /--|~~/.test($6), /!/.test($6), false, true
         # or named with `function`.
         o 'FUNCTION CALL( ArgList OptComma )CALL Block' -> (Fun $3, $6).named $1
         o 'GENERATOR CALL( ArgList OptComma )CALL Block'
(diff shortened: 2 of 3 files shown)


isiahmeadows:
How would that work with libraries using that identifier? I'm working on one that could implement this general functionality at the library level, possibly more optimally than engines could. (I'm using Bluebird-style shortcuts to do it, including keeping the stack minimized.)

summivox:
`['ID', 'async']` is left alone unless it is followed by a `PARAM(` or `FUNCTION`

summivox:
@vendethiel decided in #752 that lsc should not transpile

darky:
async/await almost in v8 as stable feature. Any activity here?

isiahmeadows:
@gkz? IIRC this barely missed the V8 train for Node 7, and it's been shipping in Edge for a while.

summivox:
My opinion:

- I don't think the syntax and semantics of async/await *in ES7* would change in any major way, now that it is being rolled out experimentally.
- The proposed syntax for LSC does not conflict with existing code, and is intended for early adopters specifically targeting ES7 (and quite possibly including further transpiling toolchain)

Therefore I believe it is safe for this to land in LSC.

isiahmeadows:
WRT the PR itself, I feel having both `(arg) ->>` (and friends) and `async (arg) ->` is rather redundant.

isiahmeadows:
@summivox You mean ES 8?

summivox:
@isiahmeadows It seems that people have stopped caring about the "version number" and started to rely on shipping schedule instead. I'm fine with this ;)

Regarding syntax: Historically coffee-ish languages have this dilemma of symbols vs English words. `&&` vs `and` in LSC is a prime example --- different operator precedence (with `and` being more sane therefore the default). I don't see why both versions of arrows cannot co-exist. However, if only one is to stay, I would suggest  `->>` which stays in line with the "symbolic" feel of the arrows while retaining orthogonality, and `async function` for the "English" side.

isiahmeadows:
There is a push to get rid of many of the duplicate operators like `&&` vs
`and` in one of the issues here. I can't remember the number right off,
though.

On Tue, Oct 25, 2016, 17:00 Yin Zhong <notifications@github.com> wrote:

@isiahmeadows <https://github.com/isiahmeadows> It seems that people have
stopped caring about the "version number" and started to rely on shipping
schedule instead. I'm fine with this ;)

Regarding syntax: Historically coffee-ish languages have this dilemma of
symbols vs English words. && vs and in LSC is a prime example --- different
operator precedence (with and being more sane therefore the default). I
don't see why both versions of arrows cannot co-exist. However, if only one
is to stay, I would suggest ->> which stays in line with the "symbolic"
feel of the arrows while retaining orthogonality, and async function for
the "English" side.

—
You are receiving this because you were mentioned.


Reply to this email directly, view it on GitHub
<https://github.com/gkz/LiveScript/pull/836#issuecomment-256174348>, or mute
the thread

summivox:
I just tested this with node v7.3.0 . **It feels awesome.**

@isiahmeadows : I agree with you on removing `async (args) -> body` now. For anonymous functions, LSC allows only arrows while ES6 allows both `function()` and arrow. There is no good reason to keep an alternative syntax for async.

dk00:
`await*` have been [removed](//stackoverflow.com/a/35168428/4578017) from [the proposal](/tc39/ecmascript-asyncawait), should we remove it as well, or simply compile it to `await promise.all ...` ?

gkovacs:
Thank you for this excellent patch. I have noticed a bug when using it against the master branch ( https://github.com/gkz/LiveScript/commit/47ad4eba3a23003b4d571ac6d7e9b0c603c8950e )

```
getdict = (name) ->>
  data = await getCollection(name)
  return {[key, val] for {key, val} in data}
```

somehow gets compiled to (notice the superfluous yield*)

```
(function(){
  var getdict, out$ = typeof exports != 'undefined' && exports || this;
  out$.getdict = getdict = async function(name){
    var data, key, val;
    data = (await getCollection(name));
    return (yield* (function*(){
      var i$, ref$, len$, ref1$, resultObj$ = {};
      for (i$ = 0, len$ = (ref$ = data).length; i$ < len$; ++i$) {
        ref1$ = ref$[i$], key = ref1$.key, val = ref1$.val;
        resultObj$[key] = val;
      }
      return resultObj$;
    }()));
  };
}).call(this);

```

summivox:
@gkovacs : this is a separate issue: #764 #938

gkovacs:
Thank you for the fast response! Also here's another issue. The following code compiles:

```
foo = -> ''

a = foo(async ->
  ''
)
```

And so does this:

```
foo = -> ''

a = foo ->>
  ''
```

But the following does not:

```
foo = -> ''

a = foo async ->
  ''
```

So in the example above, `->>` and `async ->` aren't quite identical.

isiahmeadows:
Here's my proposed resolution: remove `async ->` altogether. It doesn't fit with every other anonymous function syntax.

I'm kind of 1% tied to this project now, since the parser bugs have pretty much killed this project for me, and no stable release yet has provided ES6 support.

summivox:
I want to remove `async ->` syntax so this is a wontfix.
I just haven't been actively working on this project for a long time.

summivox:
Hmm done.

gkovacs:
I fixed the issue of yield* statements being generated inside async functions due to bug https://github.com/gkz/LiveScript/issues/764 in https://github.com/gkovacs/livescript-async/commit/97b86df09d66ea4c7e2f44d2b77ba6b6d4a3b130 (also removed `await all` as `await*` is no longer part of ES2017)

yizhong-ms:
@gkovacs could you please make a PR to my branch; thanks

gkovacs:
@yizhong-ms @summivox created a PR to your branch at https://github.com/summivox/LiveScript/pull/2/files

determin1st:
can you post some examples of this functionality?

determin1st:
does this update work? tried to run examples on livescript.net with no luck.

isiahmeadows:
@determin1st This PR hasn't been merged, and livescript.net isn't updated until the next release IIRC.

determin1st:
oh, i see. experimented with promises a little, found it useful but they are not "real" part of the language. it's like helpers. probably wont use this functionality.

gkovacs:
@rhendric Given that the ES2017 standard (with async functions) has now been released and nodejs and all major browsers support the syntax by default, is there any chance you could review and merge this pull request? Thank you!

rhendric:
I would be very happy to look it over! @summivox, can you confirm that this is in a state that you endorse for merging, or if there are any outstanding issues or design questions, summarize what they are?

summivox:
@rhendric Let me quickly build and verify. I think tests need to be added.

rhendric:
Yes please to tests. No rush from me (I'm actually running out the door right now and probably won't have free time for a few days anyway); just drop me a mention when you're ready for a review and I'll follow up when I'm able. Thanks!

willin:
mark, pls add in asap

willin:
can it use now?

gkovacs:
@willin If you're impatient and want to use it now do `npm install -g livescript-async` ( https://www.npmjs.com/package/livescript-async - it is just standard livescript plus this patch, uploaded by me to NPM as I grew impatient for this to be released). But no it is not yet released in the official livescript.

willin:
it will be great helpful if you add a documentation for this package.

ozra:
@gkovacs - thanks for the npm! Trying it out right now.

gkovacs:
I've added tests in a new branch https://github.com/gkz/LiveScript/pull/978 - as summivox asked this to be taken over in https://github.com/gkz/LiveScript/issues/977 I will be taking over the efforts for getting this merged

rhendric:
#978 has landed, so I'm closing this.

PR #837, opened 2016-02-01
support for <expr> by <expr> loops 

rhendric:
It was already in the grammar, but the lexer was written such that it could never happen.

As a special case, `for by in foo` still treats `by` as an identifier rather than a keyword.


vendethiel:
Can you please add a test with said special case? I'll merge it after that :)

rhendric:
It already exists! `test/loop.ls`, line 168

vendethiel:
Ah ok, you merely added it for that.

vendethiel:
Thanks again!

PR #841, opened 2016-02-05
add robustness to generated loop code

rhendric:
Certain expressions—integers (with or without a preceding +/-), null,
and void—when used as sources of for loops resulted in illegal
JavaScript being generated. While (42).length is a silly but legal thing
to do, 42.length is an actual SyntaxError in JS. This commit adds checks
to wrap potentially problematic expressions in parentheses.

### DIFF
--- a/src/ast.ls
+++ b/src/ast.ls
@@ -199,11 +199,13 @@ SourceNode::to-string = (...args) ->
         if once then [sub, ref <<< {+temp}] else [sub, ref, [ref.value]]
 
     # Compiles to a variable/source pair suitable for looping.
-    compile-loop-reference: (o, name, ret) ->
+    compile-loop-reference: (o, name, ret, safe-access) ->
         if this instanceof Var   and o.scope.check @value
         or this instanceof Unary and @op in <[ + - ]> and -1/0 < +@it.value < 1/0
         or this instanceof Literal and not @is-complex!
-            return [@compile o] * 2
+            code = @compile o, LEVEL_PAREN
+            code = "(#code)" if safe-access and this not instanceof Var
+            return [code] * 2
         asn = Assign Var(tmp = o.scope.temporary name), this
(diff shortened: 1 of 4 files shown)


src/ast.ls
```diff
@@ -2350,6 +2350,9 @@ class exports.For extends While
                 then "#idx #{ '<>'char-at pvar < 0 }#eq #tvar"
                 else "#pvar < 0 ? #idx >#eq #tvar : #idx <#eq #tvar"
         else
+            if @source instanceof Literal and @source.value is /^[0-9]+$/
```
vendethiel: This seems a bit too specific

rhendric: I thought so at first too, but all of the following are handled gracefully by JavaScript:
* 1e8
* 0.0
* true/false/null

This is the most general case I could find that triggers the syntax error. (Ah crap, I just found one more: negatives. I'll throw a `-?` in there.)

vendethiel: (JS doesn't even have negative literals)

rhendric: And yet `-42.length` triggers a syntax error anyway.

I'm starting to have misgivings about this approach; maybe I should be adding robustness (generating `(42).length`) instead of trying to catch the problem cases and raise errors? I was reluctant to do that at first because errors seem better than compiling things that I know are wrong, but this is turning into a slightly awkward patch. Do you have an opinion?

vendethiel: Not really, but I never had this kind of error

rhendric:
(For why-the-heck-does-he-want-this background: I was working on #831 and kept getting very annoying-to-diagnose errors from the test suite. I had to binary search the files with comments, because of course the JavaScript SyntaxErrors didn't give me a line number. Not fun. I'm just trying to help out the next guy who goes experimenting with loop, slice, or spread syntax.)

PR #848, opened 2016-02-10
fix keyword literals in object slices

rhendric:
Prior to this commit, a keyword literal such as `arguments` was treated the same way as a literal like `0` when creating object slices—so instead of `x{arguments}` meaning `{arguments: x.arguments}`, you got `{arguments: x[arguments]}`.


vendethiel:
Thanks!

PR #849, opened 2016-02-10
fix semiautovivification with slices

### LINKED ISSUE #847: Semiautovivification with slices is broken
```ls
arr = []
x = arr{}[0, 1] #=> expected [{}, {}], got {}
arr             #=> expected [{}, {}], got [null, null]
```
The above compiles to
```js
var arr, x, ref$;
arr = [];
x = (ref$ = {}, arr[0] || (arr[0] = ref$[0]), arr[1] || (arr[1] = ref$[1]), ref$);
arr;
```
but I should get something more like
```js
var arr, x, ref$;
arr = [];
x = [(ref$ = arr)[0] || (ref$[0] = {}), ref$[1] || (ref$[1] = {})];
arr;
```
right?

rhendric:
Fixes #847

### DIFF
--- a/src/ast.ls
+++ b/src/ast.ls
@@ -639,8 +639,6 @@ class exports.Chain extends Node
                 @head.called = true
             else if not @tails.1 and it.key?name is \prototype
                 @head.sproto = true
-        else if delete it.vivify
-            @head = Assign Chain(@head, @tails.splice 0, 9e9), that!, \= \||
         else if it instanceof Call and @tails.length is 1
         and bi and bi.op in logics = <[ && || xor ]>
             call = it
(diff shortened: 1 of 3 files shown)


src/ast.ls
```diff
@@ -837,6 +836,13 @@ class exports.Chain extends Node
                     i = 0
                 call <<< method: \.apply, args: [ctx or Literal \null; JS args]
 
+    expand-vivify: (o) !->
+        {tails} = this
+        i = 0
+        while i < tails.length when delete tails[i++].vivify
+            @head = Assign Chain(@head, tails.splice 0, i), that!, \= \||
+            i = 0
```
vendethiel: why not just `i--`?

rhendric: `tails` can contain multiple things in a row that aren't vivifications. The `when` guard makes `i` keep incrementing until it finds a vivification, at which point the loop body `splice`s out everything in `tails` up to that vivification into the `Assign`. `i` needs to be reset back to `0` at that point because `i` elements have been removed from `tails`.

vendethiel: Ah, duh. I use splice so much I can't even remember its arguments.. Sorry :/.

vendethiel:
Thanks!

PR #851, opened 2016-02-16
Enable better Browserify detection.

isiahmeadows:
It makes the JS compiler much easier to use with Browserify CDNs like wzrd.in, brcdn.org, etc. It shouldn't affect browser usage for existing consumers beyond what would already throw errors normally.

(Helpful for testing in browsers, where you don't have to create your own LS bundle manually just to use this feature).

Do note that this doesn't clean up all the other remaining references lying around everywhere.

### DIFF
--- a/package.json.ls
+++ b/package.json.ls
@@ -28,6 +28,7 @@ files:
   'LICENSE'
 
 main: './lib/'
+browser: './lib/browser.js'
 bin:
   lsc: './bin/lsc'
 


isiahmeadows:
@gkz I *think* this also fixes an inconsistency between the API docs and actual behavior for the browser as well. Previously, `require("livescript")` brought the Node version for both Node and the browser, but the API docs on livescript.net partially implied browsers would use `require("livescript/lib/browser")` instead, due to poor word choice.

vendethiel:
Thanks!

PR #859, opened 2016-02-21
extend cascade comprehension to support `[.. for from i to j by k]`

summivox:
This used to throw "stray ref" error.


vendethiel:
Thanks!

Issue #821, opened 2016-01-23
Consider utilizing the babel ecosystem


AriaMinaei:
LiveScript is Beautiful™. Babel otoh, has a strong ecosystem. I think there is a great opportunity here for LiveScript to have both.

What if LiveScript generated a babel-compatible AST? It would still have its own separate parser, but after parsing, it would produce an initial AST, apply a few LiveScript-specific transformations on it to preserve some semantics, and then allow all the existing babel transformers to take care of the rest.

This would allow LiveScript users to take advantage of most of the great tools that have been built around babel, namely:

* [ESLint](https://github.com/eslint/eslint): ESLint already allows custom parsers. [babel-eslint](https://github.com/babel/babel-eslint) converts babel's ASTs to that of ESLint's, which allows babel users to keep using babel's new features along with ESLint. LiveScript can do that too, and with a babel-compatible AST, that would be easy. We'll get a linter for free. There are also tools built around ESLint. People are already familiar with it. Why re-invent the wheel?
* [react-transform-hmr](https://github.com/gaearon/react-transform-hmr): Hot realoading for react components. While there is a [webpack](https://github.com/gaearon/react-hot-loader/) loader that works with LiveScript, that loader is deprecated, and to use the new one, you have to use babel. If you are using HMR, I'm sure you can't imagine going back :)
* [Babel Typecheck](https://github.com/codemix/babel-plugin-typecheck): Understands flow type annotations and adds runtime type guards.
* Other useful transforms and tools: [react-transform-catch-errors](https://github.com/gaearon/react-transform-catch-errors), [React constant elements transformer](http://babeljs.io/docs/plugins/transform-react-constant-elements/), [React display name transformer](http://babeljs.io/docs/plugins/transform-react-display-name/), [Documentation.js](https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/user-handbook.md#documentation), [astexplorer](https://astexplorer.net), ...

Note that the benefit doesn't just come from babel's existing tools. It also comes from all the activity that's been done around babel. For example, there is a lot of work put into writing [this handbook](https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/plugin-handbook.md) which explains to beginners like me, how to write custom transformers for babel. It's easy and approachable. There is no need to redo this work all over again to write the same thing for LiveScript.

[...]

PR #862, opened 2016-02-21
Start work on the babel pipeline

AriaMinaei:
As discussed in #821, these are the minimal changes required to start working on the babel pipeline.

To test it out:

``` livescript
LiveScript = require \livescript

code = """
function suma a, b
  a + c
"""

console.log LiveScript.babelPipeline.compile code
```

For now, it can only compile this code:

``` livescript
function add a, b
  a + c
```

Here is how the pipeline will probably work:
1. [`simplifyLsNode`](https://github.com/AriaMinaei/LiveScript/blob/ebd838c74ce131186614034bf577b480a2dd059a/src/babel-pipeline.ls#L13) makes LS AST easier to read with `console.log()`. This doesn't affect the rest of the pipeline, and only makes debugging easier.
2. The simplified AST is fed to [`lsASTToBabelAST`](https://github.com/AriaMinaei/LiveScript/blob/ebd838c74ce131186614034bf577b480a2dd059a/src/babel-pipeline.ls#L39). This will turn a simplified LS AST to a babel-compatible AST, similar to what `babylon.parse` returns.
3. At this point, we do have an AST that we can generate JS code from, but LiveScript's semantics, such as last-expression-gets-returned, aren't preserved. We'll need to write a bunch of babel transforms to preserve those semantics.
4. Once we have a valid babel-compatible AST with LiveScript's semantics, we can either generate JS code, like [this](https://github.com/AriaMinaei/LiveScript/blob/ebd838c74ce131186614034bf577b480a2dd059a/src/babel-pipeline.ls#L109), or feed the babel-compatible AST to babel-eslint, by changing [this line](https://github.com/babel/babel-eslint/blob/master/index.js#L6) from `var parse = require("babylon").parse` to `var parse = require('livescript').babelPipeline.parse`.


auvipy:
cool

morningrat:
No

darrrk:
I think need to create drop-in replacement of ls-ast in favor to babylon-ast.
That is:

    ls-tokens -> babylon-ast

instead

    ls-tokens -> ls-ast -> babylon-ast

Must be more faster, flexible and robust.

vendethiel:
no, that'd leave the burden of transformating Ls constructs to the lexer/grammar, which is a bad idea. It's probably better to have a "toBabelNode" or something on the AST directly (exactly in the same way we recursively compile nodes to js fragments)

JoyKrishnaMondal:
babel sucks - enuf said.

vendethiel:
@JoyKrishnaMondal That's both inconstructive and uninteresting. Let's avoid this kind of comments.

JoyKrishnaMondal:
Fine, 

Babel 6 requires a .babelrc file ( dafaq ? you guys are just a compiler - stop trying to make a platform happen - its not going to happen ) 

Babel documentation is a lot less helpful then the community is ready to admit - half the stuff is not updated to reflect recent changes. 

ES6 exists on a much lower level in terms of programming language compared to livescript - why it gets more attention is beyond my comprehension. 

Babel is more concerned about marketing then improving docs or the code itself. 

The babel community is much more dismissive then any other OSS community I have seen out there - I personally just follow their issues rather than use babel myself.

these are just some of the problems I have with babel. I just do not want babel to dominate and replace typescript,live-script,coffeescript . . .  just because they have a better marketing team.

AriaMinaei:
@JoyKrishnaMondal This better be discussed in #821. Please read my first post where I explain why I think we should this.

AriaMinaei:
I like @vendethiel's idea that we just add a `::toBabelNode()` method to each `Node` class. But I don't know if it'll be cleaner and more maintainable than `lsASTToBabelAST`. Let me do a little experiment and let you know how it goes.

vendethiel:
`lsASTToBabelAST` is basically a huge switch/case, whereas our current approach in the codebase is to use OO.

tomByrer:
Any progress?

auvipy:
really looking forward to see this moving ahead

determin1st:
@AriaMinaei 

it's a lot of work to sync those, but benefits are not clear. intellisense?

AriaMinaei:
Closing in favor of [livescript-next](https://github.com/dk00/livescript-next)

PR #863, opened 2016-02-27
Reform operator spread

rhendric:
Per conversation in #831, these commits make breaking changes to operator spreading.

* Operator spreading now requires a splat: `typeof ...[0]` => `[typeof 0]`
* Operator spreading now supports objects: `typeof ...{a: 0}` => `{a: typeof 0}`
* The LHS of assignment operators can also be spread: `++...a{b, c}` can be written as `...a{b, c} += 1`

`foo ...+...a<[b c]>` => `foo.apply(null, [+a['b'], +a['c']])` works.
`(...a=b) ->` never really worked (an array, even empty, will never be nullish) but is now an explicit error.

Close #831


vendethiel:
That looks good! It might take me a few days before I can actually review it though.
I'd gladly hear anyone's opinions on this PR. Especially if someone happens to find a case where syntaxes conflict :).

I might also want to add a few more tests. It's a pretty big change.

When merged, I'll also create an issue to track the docs state about this.

isiahmeadows:
I like the DRY aspect of this. And I can't think of any way this could currently conflict with standard argument spreading. And +1 for making odd edge cases explicit errors.

vendethiel:
Seems like no one objected, so I'm just gonna merge it :). Thanks.

Issue #857, opened 2016-02-20
Better handling of splat func arguments


summivox:
[According to Petka](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments), calling `slice$` on `arguments` kills v8 optimization. He suggested to do a naive indexed for loop on `arguments` instead.

Currently:
```livescript
f = (a, ...b, c) ->
```
compiles to:
```javascript
f = function(a){
  var i$, b, c;
  b = 1 < (i$ = arguments.length - 1) ? slice$.call(arguments, 1, i$) : (i$ = 1, []), c = arguments[i$];
};
```

summivox:
I did a patch (summivox/LiveScript@e6591cf) but it currently relies on #859 being merged. It can be done without #859 but I'm too lazy :wink:

This patch itself does not change any semantics -- the only change is inlining slicing of `arguments`. Therefore I did not add/change any tests.

isiahmeadows:
Ping @gkz (Babel and TypeScript do this by default, and [CoffeeScript has been annoying people with the slow progress](https://github.com/jashkenas/coffeescript/issues/3274).

summivox:
@isiahmeadows : currently @vendethiel  is in charge (and yes he did comment on that issue too).

BTW: #859 has been merged, so I can actually make a pull request now.

isiahmeadows:
@summivox Okay. I didn't know this.

(I don't keep up with this as much as I used to. LS 2.0 does seem pretty interesting, though.)

Issue #856, opened 2016-02-20
Superfluous closure for returned list comprehension


summivox:
```livescript
f = (a) ->
  x = [.. for a]
  y = [.. for a]
```
currently compiles to:
```
f = function(a){
  var x, res$, i$, x$, len$, y;
  res$ = [];
  for (i$ = 0, len$ = a.length; i$ < len$; ++i$) {
    x$ = a[i$];
    res$.push(x$);
  }
  x = res$;
  return y = (function(){
    var i$, x$, ref$, len$, results$ = [];
    for (i$ = 0, len$ = (ref$ = a).length; i$ < len$; ++i$) {
      x$ = ref$[i$];
      results$.push(x$);
    }
    return results$;
  }());
};
```

Can we kill that useless closure?

summivox:
[This test][test]

```livescript
# Closure-wrapped comprehensions that refer to the "arguments" object
expr = ->
  result = [item * item for item in arguments]

ok expr(2, 4, 8).join(' ') is '4 16 64'
```

Again it doesn't matter whether it is compiled to a closure or not.

BTW: leaking `arguments` is a bad idea anyway (#857).

[test]: https://github.com/gkz/LiveScript/blob/master/test/loop.ls#L389

summivox:
I realized that this is fundamental to how "level" is handled.
Never mind (for now).

PR #868, opened 2016-03-21
Better handling of splat func arguments (prevent v8 deoptimization)

summivox:
see #857, jashkenas/coffeescript#3274 for discussion


vendethiel:
Do we have a test for cases such as `(a, ...[x]:b, c) ->`?

src/ast.ls
```diff
@@ -1641,14 +1641,17 @@ class exports.Assign extends Node
         else if (ret or len > 1) and (not ID.test rite.to-string! or left.assigns rite.to-string!)
             cache = sn(this, (rref = o.scope.temporary!), " = ", rite)
             rite  = rref
-        list = @"rend#{ left.constructor.display-name }" o, items, rite
+        if rite.to-string! is \arguments and not ret
+            arg = true
+            if left not instanceof Arr then @carp 'arguments can only destructure to array'
```
vendethiel: please newline that if

src/ast.ls
```diff
@@ -1641,14 +1641,17 @@ class exports.Assign extends Node
         else if (ret or len > 1) and (not ID.test rite.to-string! or left.assigns rite.to-string!)
             cache = sn(this, (rref = o.scope.temporary!), " = ", rite)
             rite  = rref
-        list = @"rend#{ left.constructor.display-name }" o, items, rite
+        if rite.to-string! is \arguments and not ret
+            arg = true
```
vendethiel: i'd rather have a more meaningful name :)

src/ast.ls
```diff
@@ -1930,7 +1946,7 @@ class exports.Fun extends Node
             # `(a = x) ->` => `(a ? x) ->`
             else if p.op is \=
                 params[i] = Binary (p.logic or \?), p.left, p.right
-        # `(a, ...b, c) ->` => `(a) -> [[] ...b, c] = @@`
+        # `(a, ...b, c) ->` => `(a) -> [[] ...b, c] = &`
```
vendethiel: good catch

src/ast.ls
```diff
@@ -1939,30 +1955,28 @@ class exports.Fun extends Node
             params.0 = Var \it if body.traverse-children -> it.value is \it or null
         names   = []
         assigns = []
-        if params.length
-            dic = {}
```
vendethiel: was this variable unused?

src/ast.ls
```diff
@@ -1939,30 +1955,28 @@ class exports.Fun extends Node
...
-                    assigns.push Assign vr, switch
-                        | df        => Binary p.op, v, p.second
-                        | has-unary => fold ((x, y) -> y.it = x; y), v, unaries.reverse!
-                        | otherwise => v
-                    vr = v
-                else if df
-                    assigns.push Assign vr, p.second, \=, p.op, true
-                names.push (scope.add vr.value, \arg, p), ', '
+        for p in params
+            vr = p
```
vendethiel: if you want to use this occasion to give more meaningful names to the variables around, feel very free to...

src/ast.ls
summivox: I believe so. I did not find a reference in this function.

summivox:
Ah you got me. A closure is generated in the case of `...[x]:b`.
I don't have time to fix it now though...

src/ast.ls
summivox: better (refactorable) names on the way; I would rather not push before I fix the named nested splat though.

vendethiel:
That's fine, don't rush it! Just wanted to make sure it worked.

summivox:
I have a simple patch now but it is inelegant. It all boils down to this:
```livescript
[a]:b = [.. for til 4]
```
At top level this should not compile to a closure, but it does right now, i.e. #856 (funny I raised this one too).

vendethiel:
> Do we have a test for cases such as (a, ...[x]:b, c) ->?

@gkz is that okay?

Issue #854, opened 2016-02-18
Inconsistent array casting


Announcement:
# Short arrays aren't enough to loop

`[to 10]` does what expected and so does `[.. for [0 to 10]]`

## But in contrast...

`[.. for [to 10]]`  throws 
> Parse error on line 1: Unexpected 'FROM'

summivox:
Just FYI: `[.. for to 10]` now works thanks to #859

rhendric:
`[.. for to 10]` works, but `[.. for [to 10]]` still doesn't; shouldn't this remain open?

summivox:
Lexer is to blame.

summivox:
I am going to patch this temporarily with a grammar fix.

Issue #818, opened 2016-01-19
Issue warning if keyword is used as a variable name


atifaziz:
LiveScript (1.4 as of this issue reporting) allows use of keywords as variable names so to not allow escaping seems like a half-baked cake. According to @vendethiel in #816, “[Don't use to as a variable name is the usual advice](https://github.com/gkz/LiveScript/issues/816#issuecomment-172679222).” This should clearly written on [the wall](http://livescript.net/) but that's not good enough for two reasons:

1. The example in #816 shows that the program semantics can change *silently*.
2. The programmer is not protected against future keywords being introduced in the language. Like in 1, the worst case is that it may change the semantics of the program *silently*. 

The compiler should issue a warning if a keyword is used as a variable name. I'm hoping issuing a warning is the cheapest solution to implement so this will get done. :) If a new keyword is introduced later in LiveScript then the warning will help the programmer review new potential conflicts. Of course, the cherry on top would be to allow escaping so that the programmer has the choice to address the warning by escaping or renaming.

vendethiel:
I'm going to mark this as a bug for now

PR #870, opened 2016-03-22
Temporary fix for #854

summivox:
`[.. for [to 10]]` is wrongly rewritten by lexer into `[.. for [from 0 to 10]]` (not quite, see below). Patching the lexer could be non-trivial as said rewriting is currently ad-hoc (`forange` in `lexer.ls`, see #854 for discussion). This temporary solution relaxes the grammar and allow the malformed lexer output.

Caveat: `[from x to y]` is *not* supported directly from source despite the change in grammar due to the same lexer hack. The `from` is currently considered a literal rather than a `FROM` token.

A possibly cleaner solution is to get rid of `forange` and support `from` elision in a separate lexer rewrite step or in parser instead.


src/grammar.ls
```diff
@@ -422,7 +426,7 @@ bnf =
         o 'FOR ID FROM Expression TO Expression BY Expression'
         , -> new For kind: $1, index: $2, from: $4, op: $5, to: $6, step: $8
         o 'FOR FROM Expression TO Expression BY Expression'
-        , -> new For kind: $1,            from: $3, op: $4, to: $5, step: $6, ref: true
+        , -> new For kind: $1,            from: $3, op: $4, to: $5, step: $7, ref: true
```
vendethiel: Not sure why this commit isn't considered merged by github...

vendethiel:
This is far too many FOR variations... It should be refactored with some OptBy/... - also, the change should probably allow us go get rid of forange completly. That **should not** be in the lexer.

summivox:
Agreed.

There is one potential show-stopper though: In current specification `from`, `to`/`til`, `by` are so-called "contextual keywords". This makes it difficult to detect literal ranges without having the lexer do way more parsing than it should. For example:
```livescript
x = [(to til) to (by from)]
```
which is sick :joy: but sufficiently points out that "contextual keywords" are neither well-defined nor well-implemented. 

My opinion is to make at least `to`, `til`, and `by` reserved (since we are clearly talking about LS2.0 here we can stop worrying about sick cases). There has been quite some discussion on keyword usage (e.g. #818) and the conclusion seems to favor "escape yourself if you know what you're doing, otherwise don't even bother".

vendethiel:
The example you linked isn't exact an issue, because the per-nesting flags should deal with that.

But yes. let's make them reserved.

vendethiel:
@summivox would you mind making them reserved (should be easy, just need to add them to the keywords array), or do you want me to do it?

summivox:
I'm busy with my masters thesis these days. Could you please do it?

PR #884, opened 2016-04-20
Auto parse json.ls with lsc

nekdolan:
This change lets you auto compile required json.ls files in ls scripts with the node-livescript executable.

### DIFF
--- a/src/node.ls
+++ b/src/node.ls
@@ -29,7 +29,11 @@ module.exports = !(LiveScript) ->
     LiveScript <<<< events.EventEmitter.prototype
 
     require.extensions.'.ls' = (module, filename) ->
-        js = LiveScript.compile (fs.read-file-sync filename, 'utf8'), {filename, +bare, map: "embedded"} .code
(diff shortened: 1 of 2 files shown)


src/node.ls
```diff
@@ -29,7 +29,11 @@ module.exports = !(LiveScript) ->
     LiveScript <<<< events.EventEmitter.prototype
 
     require.extensions.'.ls' = (module, filename) ->
-        js = LiveScript.compile (fs.read-file-sync filename, 'utf8'), {filename, +bare, map: "embedded"} .code
+        file = (fs.read-file-sync filename, 'utf8')
+        if (filename.substr -7) is 'json.ls'
```
vendethiel: 1) that's not our codestyle
2) doesn't work with `foobarjson.ls`.

nekdolan: @vendethiel can you give me some hints or instructions for the code style? I am a beginner when it comes to livescript and functional programming.

vendethiel: just look at the rest of the codebase

and for the second point, check one more char

PR #886, opened 2016-05-02
editorconfig: Makefile uses tabs

rhendric:


vendethiel:
er, yes. Well, not all make implementations will force you to, but better make sense.

PR #894, opened 2016-06-04
Issue 751

### LINKED ISSUE #751: Curry named functions
If you have a named function that curries (I think that's how you'd name it?), the function isn't scoped properly.

    :abc (x,y) --> x + y

Expected:

	var abc;
	abc = (curry$(function(x, y){
	  return x + y;
	}));

Actual:

	(curry$(function abc(x, y){
	  return x + y;
	}));

misterfish:
This is intended to resolve #751. It doesn't provide the semantics the OP asked for but simply throws a compile-time error.

Comments very welcome of course; this is my first contribution.


vendethiel:
That looks good to me! @gkz Do you agree to throwing an error here?

gkz:
Seems reasonable, thanks

PR #732, opened 2015-05-24
repl: autocomplete globals

raine:
changes variable names to kebab-case when autocompleting

fixes #687

Issue #687, opened 2015-03-19
lsc REPL needs improved tab completion


kflu:
Both `node` and `coffee` REPLs have better tab completion than lsc. For example `glob[tab]` would be completed to `global`. `req[tab]` completes to `require`. I don't know how `coffee` implements its REPL but it feels as derived from `node`. Could `lsc` do something similar?

chisui:
this would be really helpfull

raine:
> I don't know how coffee implements its REPL but it feels as derived from node

I checked and it does it with the `use-global` option on repl.

We have the option turned off for some reason:

https://github.com/gkz/LiveScript/blob/abcb494a50fbcce258f34b59052455da42e8aa9f/src/command.ls#L276

Perhaps because variables autocomplete in camelcase?

```
ls> my-test-var = true
true
ls> myTestVar
```

Issue #307, opened 2013-05-06
split out repl


craigyk:
can the repl code be split out so that it's it own callable module?

that way I can use something like:

``` livescript
ls = require 'LiveScript'
# ... bunch of code
ls.repl.start!
```

Would be really useful when experimenting with code or debugging, interactive break points.  Bonus: hitting ctrl-D breaks out of the repl and resumes normal control flow.

PR #928, opened 2016-10-04
Fix REPL tab completion

### LINKED ISSUE #157: Tab completion for hyphenated function names in REPL
I entered the code below in the REPL and tried to use tab completion after typing 'up-'.  This doesn't work since the function name is automatically converted to upCaseName.  This is just a feature request.  Thanks.

```LiveScript
up-case-name = (.name .= to-upper-case!)
```

### LINKED ISSUE #924: Tab completion broken in LiveScript 1.5 REPL
When I type `tab` after enter `String` in LiveScript 1.5 REPL
```
ls> String.
global           process          GLOBAL           root             Buffer
clear-immediate  clear-interval   clear-timeout    set-immediate    set-interval
set-timeout      console          module           exports          require
LiveScript       path             fs               util             say
warn             die              p                pp               ppp
```
The completion result doesn't show members of String.
in LiveScript 1.4, it works well:
```
ls> String.
String.__defineGetter__      String.__defineSetter__      String.__lookupGetter__
String.__lookupSetter__      String.__proto__             String.constructor
String.hasOwnProperty        String.isPrototypeOf         String.propertyIsEnumerable
String.toLocaleString        String.toString              String.valueOf

String.apply                 String.arguments             String.bind
String.call                  String.caller                String.length
String.name                  

String.fromCharCode          String.fromCodePoint         String.prototype
String.raw                   
```

rhendric:
Commit 40fae13 seems to have broken tab complete on chain expressions in
an attempt to make kebab-case tab completion work. This PR replaces
that implementation with a larger but more robust version that actually
works.

Additional features include supporting both kebab-case and camelCase
depending on what the user has already typed, and completion on less
JavaScript-friendly syntax patterns such as binding (.~) and
accessignment (.=).

Close #924
Close #157, incidentally.
See also #307 (can this be closed too?), #687, #732


gkz:
Thanks!

PR #929, opened 2016-10-05
Fix REPL break on improper errors

### LINKED ISSUE #733: REPL breaks on some errors
Some errors break a REPL session in such way that it can't be used anymore, arrow keys for browsing history break etc.

This is on 1.4 from npm

```
$ lsc -d
LiveScript 1.4.0 - use 'lsc --help' for more information
ls> map 1, null

/usr/local/lib/node_modules/livescript/node_modules/prelude-ls/lib/List.js:13
  for (i$ = 0, len$ = xs.length; i$ < len$; ++i$) {
                        ^
[TypeError: Cannot read property 'length' of null]
ls>


...

undefined
ls> ^[[A^[[A^[[A^[[A^[[A
```

rhendric:
This is an ugly hack, but it fixes the issue where the REPL will
sometimes stop responding to arrow keys after an error has been thrown
in user code. Turns out, this happens whenever a string (or anything
other than something with Error in its prototype hierarchy) is thrown.
The fix is to restore "raw mode" on the input TTY handle when this
situation is detected, by the time-honored method of turning it off and
on again.

There's also a (basically) one-line commit in here that cleans up a double [Function] message that has been hanging around since Coco.

Close #733

### DIFF
--- a/lib/repl.js
+++ b/lib/repl.js
@@ -206,9 +206,6 @@ function repl(o, stdin, stdout){
           replCtx._ = x;
         }
         pp(x);
-        if (typeof x === 'function') {
-          say(x);
-        }
       }
     } catch (e$) {
       e = e$;
--- a/lib/repl.js
+++ b/lib/repl.js
@@ -19,7 +19,7 @@ dasherizeVars = function(str){
   }
 };
 function repl(o, stdin, stdout){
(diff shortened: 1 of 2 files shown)


heavyk:
what happened to the full repl implementation that was there last night?

I was gonna give it a go today... looked pretty extensive

rhendric:
Almost all of that REPL code is still there; I just pulled it out of `command.ls` into `repl.ls` in another recently merged commit.

PR #942, opened 2016-12-20
Add test cases and fix implicit it in object shorthands (#899)

### LINKED ISSUE #899: 'it' is not defined when referenced only by object literals?
``` ls
->
  it.name
->
  it
  {it.name}
->
  {it.name}
```

compiles to:

``` js
(function(it){
  return it.name;
});
(function(it){
  it;
  return {
    name: it.name
  };
});
(function(){
  return {
    name: it.name
  };
});
```

the third function has no argument

Another example:

``` ls
t = ->
  (cb) ->
    it
```

``` js
var t;
t = function(){
  return function(cb){
    return it;
  };
};
```

dk00:
The issue is cause by generating wrong AST node type when expanding object shorthands.

`{it.something}` is correctly expanded to `{something: it.something}`, but node type of `it` is still `Key`, and doesn't count when finding implicit `it` or `that`.

`something: it.something` parses to:
```
  Obj
    Prop
      Key something
      Chain
        Var it
        Index .
          Key something
```

`{it.something}` should do the same, but it parses to:
```
  Obj
    Prop
      Key something
      Chain
        Key it
        Index .
          Key something
```
Changes are made to convert node type to `Var` when expanding object shorthands, so that generated AST is consistent.

### DIFF
--- a/src/ast.ls
+++ b/src/ast.ls
@@ -294,6 +294,7 @@ SourceNode::to-string = (...args) ->
     unparen       : THIS
     unwrap        : THIS
     maybe-key     : THIS
+    maybe-var     : THIS
     expand-slice  : THIS
     var-name      : String
     get-accessors : VOID
@@ -567,6 +568,8 @@ class exports.Key extends Node
     is-complex: NO
 
     assigns: -> it is @name
+
+    maybe-var: -> (Var @name) <<< {@line}
 
     var-name: ->
         {name} = this
--- a/src/grammar.ls
+++ b/src/grammar.ls
@@ -341,9 +341,9 @@ bnf =
 
     # The various forms of property.
     KeyValue:
-        o 'Key'
+        o 'Key' -> $1.maybe-var!
         o 'LITERAL' -> Prop L(1,Key $1, $1 not in <[ arguments eval ]>), L 1 Literal $1
-        o 'Key     DOT KeyBase' -> Prop $3, Chain(            $1; [L 2 3 Index $3, $2])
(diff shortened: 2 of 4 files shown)


rhendric:
Be advised, this is an incomplete fix—expressions like `-> {it=\default}` still have `Key it` in their trees. (As I mentioned in #899, I have some pending work in this area, so I might clobber this fix if it gets merged before I get back to compiler hacking.)

dk00:
Thank you.

I added the test case and checked things again,  and confirmed that both `-> {it=\default}` and `-> {it.name=\default}` are covered by this fix.

`it` is converted to `Var` when deriving from `KeyValue LOGIC Expression` to `(Key) LOGIC Expression` or `(Key DOT KeyBase) LOGIC Expression`.

rhendric:
Ah, you're right and I'm sorry! I had checked out your code and I thought I had done a `make force` before testing but I must not have; my bad. (Is there a reason you're not including the recompiled lib files in your commits?)

dk00:
Compiled & generated things are not under version control in other projects I'm participating in, and there's `git checkout -- lib` after `npm test`  ----  all compiled `lib/*.js` are reverted after testing.

I thought the revert is to make normal commits clean, and generated files are committed only when releasing a new version. Should I include these compiled files?

rhendric:
It's what gkz seems to do, and so I've always included them. Not sure why the revert is there after `npm test` though. 🤷‍ PRs seem to get merged with or without them, so... do what you think is best, I guess? But it would decrease the chances of some over-hasty commenter accidentally giving unwarranted criticism to your nice PR. 😃

gkz:
cool thanks

Issue #953, opened 2017-03-07
Wrong line number reported on runtime error


pavel:
Having a following `index.ls` file:
```livescript
for i from 0 to 2
	console.log 'Hello', ind
```

When running it with:
```bash
lsc index.ls
```

I get the following error:
```
Failed at: index.ls
ReferenceError: ind is not defined
    at Object.<anonymous> (/tmp/livescript-test/index.ls:5:24)
1| // Generated by LiveScript 1.5.0
2| var i$, i;
3| for (i$ = 0; i$ <= 2; ++i$) {
4|   i = i$;
5+   console.log('Hello', ind);
6| }
7| 
8| 
9| 
    at Module._compile (module.js:571:32)
    at Object.LiveScript.run (/tmp/livescript-test/node_modules/livescript/lib/node.js:31:19)
    at compileScript (/tmp/livescript-test/node_modules/livescript/lib/command.js:222:31)
    at /tmp/livescript-test/node_modules/livescript/lib/command.js:152:11
    at /tmp/livescript-test/node_modules/livescript/lib/command.js:142:7
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:416:3)
```

The problem is that line number `5` reported in error log for `index.ls` is incorrect in scope of `index.ls` file. In this example I get a full compiled version as (probably) the demo is short. But when working with bigger files, I do not get a full compiled version, so it becomes difficult to track errors.

I tried to use sourcemaps by running:
```bash
lsc -m embedded index.ls
```
the result is still the same.

Is there a way to make errors show on correct lines in scope of `ls` files?

------
OS: Linux
Node.js: 7.5.0
LiveScript: 1.5.0

rhendric:
One thing you could do right now is compile `index.ls` (with `-m embedded`, as you discovered) instead of running it directly from lsc, and when you run it, run it with `node -r source-map-support/register index.js` (after pulling down [source-map-support](https://github.com/evanw/node-source-map-support) from npm).

I don't have the time right this second to try integrating this functionality into the lsc command when running scripts directly, but it shouldn't be that hard if someone else wanted to try it.

PR #956, opened 2017-03-08
fix gkz/LiveScript#953

rhendric:
@pavel made a very reasonable request that lsc report correct line
position information when running LiveScript files and encountering run
time errors. This commit implements that, at the cost of a few minor
changes:

* Embedding a source map is now the default behavior for lsc when
  running code instead of just compiling.

* The feature where the generated JavaScript source is woven into the
  stack trace is disabled when source map information maps a stack frame
  back to LiveScript. It would perhaps be nice, but a larger change, to
  weave in LiveScript source in the same way in that case.  But there
  are other tools that can do this when given correct frame source
  information as will be produced now, not to mention it's easy to look
  in the relevant file manually.

* Source maps and the sourceMappingURL directive now correctly refer to
  each other using just basenames. This should make anything consuming
  LS source maps happier, not sadder, since the standard is for relative
  paths in source maps to be interpreted relative to the directory of
  the referring file.


gkz:
cool thanks

Issue #941, opened 2016-11-20
Use of splats in object destructuring


lorefnon:
Current behavior when using splats while destructuring objects is non-intuitive (if not buggy): 

```livescript
x = ({name, age, ...rest}) -> 
  rest
```

Results in: 

```javascript
var x;
x = function(arg$){
  var name, age, rest;
  name = arg$.name, age = arg$.age, rest = import$({}, arg$.rest);
  return rest;
};
function import$(obj, src){
  var own = {}.hasOwnProperty;
  for (var key in src) if (own.call(src, key)) obj[key] = src[key];
  return obj;
}
```

People coming from ES6 are likely to expect rest to include attributes other than name and age. 

Analogous example using babel: 

```javascript
var x = ({a, b, ...rest}) =>
  console.log(rest)
```

compiles to:

```javascript
"use strict";

function _objectWithoutProperties(obj, keys) { var target = {}; for (var i in obj) { if (keys.indexOf(i) >= 0) continue; if (!Object.prototype.hasOwnProperty.call(obj, i)) continue; target[i] = obj[i]; } return target; }

var x = function x(_ref) {
  var a = _ref.a,
      b = _ref.b,
      rest = _objectWithoutProperties(_ref, ["a", "b"]);

  return console.log(rest);
};
```

Is there a rationale behind the current behavior ?

rhendric:
I've scratched my head at this one before; would like to hear other opinions but personally I'd be happy to call this a bug.

summivox:
Good catch! I don't think there is any rationale for current behavior either.

Nitpick: the `_objectWithoutProperties` impl could have used an object instead of an array of strings

PR #958, opened 2017-03-16
Object syntax work

rhendric:
Okay, here's a big chunk of work all aimed at rationalizing the tangled web of features surrounding object property shorthand, object slices, object destructuring, and the undocumented ‘property logic’ feature which permits `{key: a ? 1} = b` to mean `a = b.key ? 1`, and similarly for `and`, `or`, `&&`, or `||` in place of `?`.

This PR does contain breaking changes, most of which are defensive (a thing that used to compile but with likely unexpected behavior is now an error). The exception is the behavior requested in #941, which is implemented here.

The individual commits have tests corresponding to each incremental unit of work, but here's a summary of all the language changes:

### Were broken and now work
* `o{(a.b)}` now correctly accesses `o[a.b]` instead of... wait for it... `o.ref$`
* `o{k: foo ? bar}` now is equivalent to `{k: o.foo ? bar}`, just like `o{foo ? bar}` is equivalent to `{foo: o.foo ? bar}`. `o{k: foo ? bar}` used to mean `{k: o[foo ? bar]}`.

### Were weird and now error
* `o{k: a.complex.expression}` used to mean `{k: o[a.complex.expression]}`, but `o{k: foo}` doesn't mean `{k: o[foo]}`; it means `{k: o.foo}`, so this is inconsistent. Now `o{k: a.complex.expression}` is an error. If you really want the old behavior, you can always write `o{k: (a.complex.expression)}`.
* `{a xor b}` used to parse but complain about `a` and `b` not existing. Now it throws a clearer ‘invalid property shorthand’ error.
* Sticking labels on objects and arrays outside of a pattern (`foo.bar = [a, b]:c`, as opposed to `[a, b]:c = foo.bar`) never did anything and is now a syntax error.
* `{...x ? y} = z` is now a syntax error. It used to compile to `x = ({} <<< z.x) ? y` but the default value never applied, because the LHS of `?` was always an object and thus never nully. If #941 weren't being implemented, maaaybe it would make sense to change this to mean `x = {} <<< (z.x ? y)`, but given that #941 is part of this work, even that implementation doesn't make sense, so it's an error.

### Didn't parse and now do
* `{a.b!c?key}` is now valid everywhere `{a.key}` used to be (`key` is the key in both forms).

### Outright changes to working features
* `{a, b, ...c} = a: 1 b: 2 c: 3 x: 4` now means roughly `a = 1; b = 2; c = c: 3 x: 4` instead of `a = 1; b = 2; c = {} <<< 3`; see #941. I hope we can all agree this is a much more obvious behavior.

I made an implementation decision here that may be controversial: `{a, b, ...c} = o` doesn't mean the same thing as `{a, ...c, b} = o`. In the latter, if `o.b` exists, `c.b` will also (whereas it is excluded along with `a` in the former). Basically, the splat only excludes keys that precede it. This is symmetric with the behavior of `{a, ...c, b}` in value position, where `c.a` can clobber `a`, but `b` will clobber `c.b`. This also means that it's still valid to have multiple splats in an object destructuring: in `{a, ...b, c, ...d} = o`, `b.c` may exist, but `d.c` won't. I don't know if this behavior is useful to anyone, but it's strictly more powerful than the alternative, and unlikely to surprise people who aren't looking for it, since the natural thing to do is to put the splat at the end of the object.


rhendric:
I'm a little uncomfortable just merging this myself with zero feedback on it, so I'll wait **two weeks** for comments and, if there are no objections, merge it on **May 15**.

Issue #963, opened 2017-05-06
Use Array.from instead of [].slice


KSXGitHub:
### Example code

```livescript
function * gen ()
  yield 123
  yield 456
  yield 789

console.log [...gen()]
```

### Expected JavaScript

```javascript
function* gen(){
  (yield 123);
  (yield 456);
  return (yield 789);
}
console.log(Array.from(gen())); // This would print [123, 456, 789]
```

### Current JavaScript
```javascript
var slice$ = [].slice;
function* gen(){
  (yield 123);
  (yield 456);
  return (yield 789);
}
console.log(slice$.call(gen())); // This would print an empty array
```

dk00:
[`Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) requires ES6 support, and if a JS engine knows `Array.from`, it should also know  [spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator)(`...expression`)

If we decide to output ES6, we can use spread here, spread can be used with any iterables(including generators) like `Array.from`.

```ls
splat-in-array = [a, ...b, c, ...gen!]
splat-in-object = {a, ...b, c, ...d}
call-with-splat-parameters = fn a, ...b, c, ...d
```

```js
var splatInArray, splatInObject, callWithSplatParameters
splatInArray = [a, ...b, c, ...gen()];
splatInObject = {
  a,
  ...b,
  c,
  ...d
};
callWithSplatParameters = fn(a, ...b, c, ...d);
```

rhendric:
Targeting ES6 has been discussed elsewhere, I believe, and if it's worth doing, it's worth doing in the context of a broader review of the backend, such as #862. In the meantime, this can certainly be addressed in an ES3-compatible way, which is my intent.

KSXGitHub:
@dk00 We can use [polyfill](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from?v=control#Polyfill)

vendethiel:
@rhendric congrats on the badge!

determin1st:
thanks 8)

PR #965, opened 2017-05-15
use Array.from to expand array splats

rhendric:
Prior to this commit, [].slice was used to turn array-like things into
arrays when needed for splats in array literals. This commit uses
Array.from instead, when available at runtime, and falls back to slice
when not. This allows a wider range of iterables to be used as splats,
on the platforms that support them (platforms missing Array.from will
also be missing more exotic iterables).

Fix gkz/LiveScript#963.

---

As a relatively simple bug fix, I'm setting the comment period for this at **one week**, after which I will merge on **May 22** if nobody has any concerns.


src/ast.ls
```diff
@@ -3228,6 +3229,8 @@ UTILS =
       }
     }'''
 
+    array-from: 'Array.from || function(x){return slice$.call(x);}'
```
gkz: Can the second part of "or" just be `slice$.call`? Or is there a reason for the wrapping function

rhendric: I did try that, but `Function.prototype.call` needs something to bind it to `slice$` or it forgets where it came from.

Issue #962, opened 2017-05-02
Block-scoped variables


KSXGitHub:
For example:

```livescript
test = 'foo'
console.log test # expect: foo

if true
   test = 'bar'
   console.log test # expect: bar

console.log test # expect: foo

determin1st:
JavaScript has a function scope by default. It is very useful. So "expect"ation is not correct.

KSXGitHub:
@determin1st 

1. JavaScript `var` keyword is function-scope, but `let` and `const` is block-scope

2. I am aware that LiveScript cannot enforce `var` to be block-scope. Hence, I suggest:

 * By default, LiveScript is compatible with old JavaScript standard, so it cannot use `let` and `const`, instead, it should rename block-scoped variables.

* If user tell LiveScript to not worry about old JS standard, but to worry about es6 only, `let` and `const` can be used.

determin1st:
can you make an example, where block scope has practical superiority over function scope? i never encounter such a situation.

KSXGitHub:
@determin1st I should ask you an example where local variable has practical superiority over global variables then? Because everything just fine being global, why implement scope/context at all!

ES6 implements `let` and `const` to make block-scope possible, and it do it for a reason: The JavaScript comunity demand it, and they also demand it for reasons.

determin1st:
@KSXGitHub 

yes, but it's not an example) it doesnt matter what crowd says until me see the benefits.
function scope is not global scope, so if there is a need to reduce scope, no problem to add a function inside. JavaScript is not class-typed-imperative language, it's prototypal-ducktyped-functional style.

ES6 let/const/class may give something, but i cant see what (practical application).

KSXGitHub:
@determin1st I could add a function inside and call it immediately, the problem is, it's ugly. For instance, I could add a `do ->` which then transform into JS `(function () { ... })()`. In JS, I have no problem writing that extra `(function () {` and `})()`. In LiveScript however, I have to add one extra line for `do ->` and one extra level of indent.

determin1st:
@KSXGitHub 

lot of things may be solved using both functions and fors. You say, that

a.forEach (el, index) !->
  # some code

or

for el,index in a
  # some code

is uglier than

a.forEach(function(el, index) {
  # some code
});

or

for (let index=0; index < a.length; index++) {
  el = a[index];
  # some code
}

[...]

PR #967, opened 2017-05-15
bodyless let

rhendric:
This commit adds two language features involving the `let` keyword. The first is bodyless let, which permits a `let` line without an indented body to be interpreted as implicitly containing the remainder of the current block as its body, much like backcall syntax. The second is `let const` and `let var`, which are shorthands for including const or var declarations as the first line in the body of a let expression, whether that body is indented or implicit.

Resolve gkz/LiveScript#962.

---

This is a new feature which would represent some support burden going forward (it touches the lexer, although in a straightforward way), and thus deserves some attention. I'll hold this open for at least **two weeks**, merging on **May 29** if met with only apathy (there has been a relatively spirited conversation in #962 already).


rhendric:
For anyone watching, I'm leaning towards not merging this. Per discussion in #962, I suspect using a scope-introducing `let` without the block demarcating the scope introduces more gotchas than it's worth.

vendethiel:
`let const` might be a good idea however.

rhendric:
So the question is, should `let const a = b, b = a` have the sequential assignment semantics of `const a = b, b = a`, or the swappy function argument semantics of `let a = b, b = a`?

If the former, then
```ls
let const foo = stuff
  more-stuff
```
is just shorthand for
```ls
let
  const foo = stuff
  more-stuff
```
and I don't think saving that line is especially worth it if not in service of the use case of #962. (This is what the PR implements.)

If the latter, then I think we should also be talking about consts in function parameters, like `(a, const b) -> ...`. I think this would be very interesting to pursue (if the parser ends up easily permitting it) but is nothing like what I wrote in this PR.

vendethiel:
That's a good question, But I'm not sure `const` should be the keyword to swap between sequential  and parallel assignment. Lisps usually use `let` and `let*` (which might not be clear either, but at least not as confusing as `const`).

rhendric:
Yeah, I take your point, although I wouldn't think of `const` as the thing that switches between assignment semantics. I was more thinking, do you interpret `let const a = b, b = a` as a `const a = b, b = a` inside a `let`? Or do you interpret it as sugar for the currently impossible `((const a, const b) -> ...) b, a`?

vendethiel:
The latter, like `let` behaves normally.

rhendric:
I think so too, but then I think it also behooves me to investigate allowing that syntax in functions.

PR #969, opened 2017-06-19
invert sense of ? with compound assignment

rhendric:
The code `a ?= b` is basically `a? || a = b`, which is all well and good; it makes sense to want to assign to `a` if it's nullish.

However, prior to this commit, the code `a ?+= b` was basically `a? || a += b`, which is useless. If `a` is nullish, I certainly don't want to add anything to it.

Contrast with accessignment, where `a?=b` is basically `a? && a.=b`. Obviously better this way.

This commit makes `a ?_= b` mean `a? && a _= b`, as with accessignment, for any compound assignment operator. (`a ?= b` and `a ?:= b` are unchanged.)

Technically a breaking change, but I don't see how the feature would have been put to any good use before, so I'm treating this as a minor bug fix and waiting **one week** before merging on **June 26**, unless, as always, there are objections.

### DIFF
--- a/lib/ast.js
+++ b/lib/ast.js
@@ -2600,13 +2600,17 @@ exports.Assign = Assign = (function(superclass){
     return sn.apply(null, [null].concat(arrayFrom$(code)));
   };
   Assign.prototype.compileConditional = function(o, left){
-    var lefts, morph;
-    if (left instanceof Var && in$(this.logic, ['?']) && this.op === '=') {
+    var ref$, lcache, morph;
+    if (left instanceof Var && this.logic === '?' && this.op === '=') {
       o.scope.declare(left.value, left);
     }
-    lefts = Chain(left).cacheReference(o);
+    ref$ = Chain(left).cacheReference(o), lcache = ref$[0], left = ref$[1];
     o.level += LEVEL_OP < o.level;
-    morph = Binary(this.logic, lefts[0], (this.logic = false, this.left = lefts[1], this));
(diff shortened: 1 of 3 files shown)


vendethiel:
Which case does this serve that `&&_=` doesn't?

rhendric:
`?` and `&&` differ on falsy-but-not-nullish values (0 and the empty string). So
```ls
actor.health ?+= 10
```
increments actor’s health if it exists (even if it's 0), whereas
```ls
actor.health &&+= 10
```
only does so if actor has a nonzero health.

I feel like there are at least as many, if not more, use cases for the former as for the latter.

Again, I'm shooting for consistency with what accessignment does: compare `actor.health?=add 10`, if health for some reason is an object with an add method instead of a primitive number. Forbidding this with operators but permitting it with accessignment doesn't make a lot of sense to me, though I'm open to arguments that, e.g., the operator version is more confusing for some reason.

vendethiel:
I'm not too happy on `a ?= b` being different from `a?=b` either, tbh. I know why it's different, but I think requiring an explicit `a?.=b` isn't too bad.

rhendric:
That seems like a completely independent question—or am I missing your point? Even with the explicit dot, `a ?.= b` does nothing if `a` is nullish, and I'm making the case that `a ?+= b` should similarly do nothing if `a` is nullish—basically, compound assignment operators are more like accessignment then they are like simple assignment, in that they both read and write the LHS. Any further concerns about that?

vendethiel:
No, that's a completely independent point indeed.
I'm fine with this PR as-is, I only have doubts that it won't actually be used.

PR #974, opened 2017-07-22
Change Freenode webchat with Kiwi IRC

ShalokShalom:
Its a modern alternative with image preview, tab completion and so on.


igl:
https://webchat.freenode.net/ is now using kiwi 🤷

rhendric:
So, close then? (As someone who is only on IRC maybe once a year, I don't really have an opinion one way or the other about whether this is good, bad, or unnecessary.)

Issue #986, opened 2017-11-18
`x in ["a"]` (list with one item) is not optimized


oprypin:
```livescript
x in ["a", "b"]
```
```javascript
x === "a" || x === "b";
```
---
```livescript
x in ["a"]
```
```javascript
in$(x, ["a"]);
function in$(x, xs){
  var i = -1, l = xs.length >>> 0;
  while (++i < l) if (x === xs[i]) return true;
  return false;
}
```

rhendric:
Hmm, this behavior is ancient (the culprit is commit 95f7a6856, from the distant year 2011). I don't see any reason why the one- (and zero-) element case(s) couldn't be optimized like this.

PR #987, opened 2017-11-18
optimize `x in [0]`, `x in []`

rhendric:
This applies the same optimization in use for `x in [0 1]` to one- and zero-element array literals.

Fix gkz/LiveScript#986.

---

This is a small optimization, so I'm setting the comment period for this at **one week** (plus change, to account for Thanksgiving weekend), after which I will merge on **Nov 27** if nobody has any concerns.

### DIFF
--- a/test/operator.ls
+++ b/test/operator.ls
@@ -88,6 +88,15 @@ a = [1 2 3]
 ok not ("2" in a)
 ok not ("2" in [1 2 3])
 
+# [LiveScript#986](https://github.com/gkz/LiveScript/issues/986)
+# Optimize `in` for one- and zero-element array literals
+eq "x() === 0;"    LiveScript.compile 'x! in [0]'     {+bare,-header}
+eq "x() !== 0;"    LiveScript.compile 'x! not in [0]' {+bare,-header}
+eq "(x(), false);" LiveScript.compile 'x! in [ ]'     {+bare,-header,+no-warn}
+eq "(x(), true);"  LiveScript.compile 'x! not in [ ]' {+bare,-header,+no-warn}
+eq "false;"        LiveScript.compile 'x in [ ]'      {+bare,-header,+no-warn}
+eq "true;"         LiveScript.compile 'x not in [ ]'  {+bare,-header,+no-warn}
+
 # Non-spaced values still work.
 x = 10
 y = -5
--- a/lib/ast.js
+++ b/lib/ast.js
@@ -2964,13 +2964,24 @@ exports.In = In = (function(superclass){
   }
   In.prototype.children = ['item', 'array'];
   In.prototype.compileNode = function(o){
-    var array, items, code, ref$, sub, ref, cmp, cnj, i$, len$, i, test;
+    var array, items, value, code, ref$, sub, ref, cmp, cnj, i$, len$, i, test;
     items = (array = this.array).items;
-    if (!(array instanceof Arr) || items.length < 2) {
+    if (!(array instanceof Arr)) {
       return sn(this, this.negated ? '!' : '', util('in'), "(", this.item.compile(o, LEVEL_LIST), ", ", array.compile(o, LEVEL_LIST), ")");
     }
+    if (items.length === 0) {
+      if (!o.noWarn) {
+        this.warn("value can never be `in` an empty array");
+      }
+      value = !!this.negated + "";
+      return this.item.isComplex()
+        ? sn(this, "(", this.item.compile(o, LEVEL_LIST), ", ", value, ")")
+        : sn(this, value);
+    }
     code = [];
-    ref$ = this.item.cache(o, false, LEVEL_PAREN), sub = ref$[0], ref = ref$[1];
+    ref$ = items.length === 1
+      ? [ref$ = this.item.compile(o, LEVEL_PAREN), ref$]
+      : this.item.cache(o, false, LEVEL_PAREN), sub = ref$[0], ref = ref$[1];
     ref$ = this.negated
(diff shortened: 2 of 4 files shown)


test/operator.ls
```diff
@@ -88,6 +88,15 @@ a = [1 2 3]
...
 ok not ("2" in [1 2 3])
 
+# [LiveScript#986](https://github.com/gkz/LiveScript/issues/986)
+# Optimize `in` for one- and zero-element array literals
+eq "x() === 0;"    LiveScript.compile 'x! in [0]'     {+bare,-header}
+eq "x() !== 0;"    LiveScript.compile 'x! not in [0]' {+bare,-header}
+eq "(x(), false);" LiveScript.compile 'x! in [ ]'     {+bare,-header}
+eq "(x(), true);"  LiveScript.compile 'x! not in [ ]' {+bare,-header}
+eq "false;"        LiveScript.compile 'x in [ ]'      {+bare,-header}
+eq "true;"         LiveScript.compile 'x not in [ ]'  {+bare,-header}
```
ymeine: Those two use cases can break if someone defines a property `x` on the global object with a getter having side-effects. I think the optimization should be removed.

vendethiel:
Why not error for an empty array?

test/operator.ls
rhendric: LiveScript universally (I'm pretty sure) assumes that its users won't do such things. For example, look at the two-element case for this same optimization:
```ls
x in [0 1]
```
compiles to
```js
x === 0 || x === 1;
```
which would also ‘break’ if getting `x` had a side effect by invoking it twice (compare `x! in [0 1]`, which caches a reference to `x!`). It's not just the `in` operator; many kinds of syntax sugar in LiveScript (array/object slices, destructuring, accessignment, the list goes on) make a decision to cache or not cache some expression based on whether it's presumed to possibly have side effects, and simple variable access is never so presumed.

Changing our mind on that is a bigger change than this optimization alone and would merit a bigger discussion. I'm open to that discussion happening, but I don't think it should block this PR, especially since the only non-test code change necessary would be to change the `is-complex` method which this PR uses.

@ymeine, if you're concerned about this behavior, would you start up that discussion in a new issue?

rhendric:
@vendethiel, because there's nothing ambiguous or difficult to understand or handle about the empty array case, IMO. It's a silly thing to write but so is the singleton array case, and as you can see from the files changed, there were two examples of that in our very own lexer. I figure users don't set out to write unnecessarily complex expressions; they write expressions that maybe need the full power of `in` and then add and remove elements as their code evolves. If a user wants to keep `in [x]` or `in [ ]`, for symmetry with other parts of the code, for ease of code generation, or simply out of sheer laziness, I don't see why LS should force them to change since the meaning is still perfectly clear.

Contrast with, for example, array slices, where `x[]` is an error instead of desugaring to `[]`. Here, there is a good reason to error, because there's a discontinuity in the syntax: `x[0]` is simple element access, not a one-element slice. So it's natural to wonder what the ‘correct’ next element in the sequence `x[0 1 2]`, `x[0 1]`, `x[0]`, ??? should be. Since that case is both silly *and* ambiguous, it makes sense to not define it and make it a syntax error. But `x in []` is quite unambiguous—that code has a meaning today and shouldn't change.

test/operator.ls
ymeine: Well, given the fact that: 

- first it is quite unlikely to define such a getter on the global object
- then it is quite silly to expect triggering side-effects as a side-effect of an expression itself
- finally I wouldn't write such code which doesn't make much sense except for the reasons you explained as an answer to @vendethiel 

I wouldn't fight too much for this. Also because your arguments make sense.

However I think that for the sake of consistency any LiveScript expression that involves the evaluation of an input expression should evaluate it only once and cache it, no matter how many times it is actually used afterwards to compute the result of the expression. Optimizations make things asymmetrical. Also, the distance between `x!` and `x` is very short, so expecting different behavior in those cases is somewhat disturbing.

I might open an issue later on as you suggested.

rhendric: I think we by and large agree: if you find any instances of LiveScript not compiling to code that evaluates an input expression exactly once, if that expression is anything more complex than an atomic literal or a simple variable, that's a bug. (The caching might not be necessary if the output code naturally uses the expression exactly once, unless the output code would evaluate that expression out of order.) I think the only thing we might disagree on is whether variables should be deprived of their exemption status.

ymeine: Exactly. However your choice of wording made something clear: for variables it's perfectly fine to optimize, for properties it's not. The case I highlighted was for a global property - whose access looks like a variable access - and that can be checked when compiling, maybe it even occurs / would occur in the `is-complex` method.

vendethiel:
I disagree. It's not about being "difficult to understand"-LS doesn't really care about that. It's about never meaning what you want. The only case you'd have such code is after a refactoring.

rhendric:
Hm. Well, this does give me pause—I wasn't expecting controversy here. I am worried about the idea of taking an expression that isn't confusing and used to work and making it an error without good justification, and I'm not sure that ‘you probably meant to remove this code instead’ is good enough. I'm not aware of much precedent for that level of paternalism here—even the complaint about `0 == \zero` is a warning, not an error, and IMO that's a much clearer sign of an issue with your code.

vendethiel:
A warning would be totally fine IMHO.

rhendric:
Okay, let's start this up again. Since the compiler warnings proposal didn't generate any objections, I merged it, and now this PR has warnings. I'll give anyone **one more week** to have second thoughts and then merge on **March 14**.

PR #994, opened 2018-01-10
Fix interactions of loop guards+let(+destructuring), closes #992

### LINKED ISSUE #992: When using for+let, the loop variable isn't available in the when clause
exemple:

```livescript
for let k in v when k
  ...
```

result:

```js
var i$, ref$, len$;
for (i$ = 0, len$ = (ref$ = v).length; i$ < len$; ++i$) {
  if (k) {
    (fn$.call(this, ref$[i$]));
  }
}
function fn$(k){
  throw Error('unimplemented');
}
```

expected:

```js
var i$, ref$, len$;
for (i$ = 0, len$ = (ref$ = v).length; i$ < len$; ++i$) {
  var k = ref$[i$];
  if (k) {
    (fn$.call(this, k));
  }
}
function fn$(k){
  throw Error('unimplemented');
}
```

vendethiel:


src/ast.ls
```diff
@@ -2475,11 +2474,13 @@ class exports.For extends While
         super body
 
         if @guard and @let and (@index or @item)
-            @body.lines[0].if.traverse-children !~>
+            # if @item is an Obj, then we're destructuring. Check for @item.name instead, in case of an as-pattern.
+            item-name = if @item instanceof Obj then @item.name else @item.value
+            @body.lines[0].traverse-children !~>
```
vendethiel: removed `.if` so that we also traverse the head.

rhendric:
This looks like it does the right thing for the cases you identified, but it doesn't leave me with a lot of confidence that there aren't other edge case bugs still lurking in this part of the code. Everything inside the `if @guard ...` is kind of duplicating some responsibility that already exists elsewhere (in the `if @let` above, or in `While::add-body`), and this PR adds more by mixing in a small amount of object destructuring responsibility. This isn't your fault—the PR is just adding atop badness that existed before—but IMO the best resolution of this bug would refactor the loop AST classes slightly to eliminate the need for this `if` block. Do you want to take a crack at that?

vendethiel:
I guess it's somewhat my bad for introducing poorly the if-let previously?

The while case isn't affected because we don't have the "no function in loops" interaction there. We do call super (-> While) in For.

Which if block do you want to remove?

rhendric:
Oh, ha. Then yes, j'accuse!

I was referring to the `if @guard`. What I was thinking was something along the lines of pulling out an `apply-guard` method in `While` which replaces `@body` with `If @guard, @body` and deletes `@guard`, using that in `While::add-body` and also in `For::add-body` after the `if @let` but before the `super`. I haven't tried it yet so there may be some ill effect there that I haven't considered, but if it works then the entire `if @guard` can be removed, since the guard will have been incorporated into the body already.

vendethiel:
The body is replaced in the super call already.

While doesn't have this pass, though, because "while let" makes pretty much no sense.

rhendric:
I was a little confused. I am unconfusing myself by attempting to write code.

BTW, here's one of those lurking bugs that this PR doesn't fix:
```ls
for let {k} in v when k
  ...
```
gives
```js
var i$, ref$, len$;
for (i$ = 0, len$ = (ref$ = v).length; i$ < len$; ++i$) {
  if (k) {
    (fn$.call(this, ref$[i$]));
  }
}
function fn$(arg$){
  var k;
  k = arg$.k;
  throw Error('unimplemented');
}
```

vendethiel:
Very good point. I'm thus not sure what the best course of action is....

PR #995, opened 2018-01-10
Fix flag combinations: -ce and -aj, fixes #993

### LINKED ISSUE #993: Flags getting randomly ignored
Some flag combinations broke.

------

`$ lsc -bce 'a!'`
Result: `ReferenceError: a is not defined`
Expected: `a();`

-----

`$ lsc -aje 'a!'`
Result: apparently an infinite loop?
Expected: a parse tree as json

vendethiel:
So these work again:

```
$ ./bin/lsc -m none -bce 'a!'
a();
$ ./bin/lsc -aje '1'
{
  ...
}
```

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -74,7 +74,7 @@ version = LiveScript.VERSION;
       if (positional.length && (o.json || /\.json$/.test(positional[0]))) {
         o.json = true;
         fshoot('readFile', positional[0], jsonCallback);
-      } else if (o.json) {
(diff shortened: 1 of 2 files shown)


rhendric:
The `-bce` fix looks good to me.

Prior to this, both `-aje` and `-je` read from stdin. Now `-aje` doesn't but `-je` still does. Why the difference?

rhendric:
Ah, never mind! I get it now (the thing being read in is only relevant when actually running the code).

PR #996, opened 2018-01-11
fix loop guards + let + destructuring issues

### LINKED ISSUE #992: When using for+let, the loop variable isn't available in the when clause

rhendric:
Alternative to #994, closes #992.

### DIFF
--- a/lib/ast.js
+++ b/lib/ast.js
@@ -891,8 +891,8 @@ exports.Var = Var = (function(superclass){
     return this$;
   } function ctor$(){} ctor$.prototype = prototype;
   prototype.isAssignable = prototype.isCallable = YES;
-  Var.prototype.assigns = function(it){
-    return it === this.value;
+  Var.prototype.assigns = function(){
+    return [this.value];
   };
   Var.prototype.maybeKey = function(){
     var ref$;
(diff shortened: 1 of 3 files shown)


rhendric:
@vendethiel, review?

vendethiel:
Seems to work in the cases I tested. I still would like if you added tests like:

nested (`for let [{a}] in b`)
with an alias (`for let {k: {v: a}} in b`)

rhendric:
Is this better? Any more cases you think are relevant?

vendethiel:
LGTM.

Issue #1000, opened 2018-01-18
Break listar compat to always mean array (or remove it)


vendethiel:
We really need to fix listar, it's been unintuitive for too long.

```livescript
a =
 * b
```
should not mean `a = b`.

The offending code is [here](https://github.com/gkz/LiveScript/blob/68a6d215e790eccb1047af0b7c91f9b695adadc6/src/lexer.ls#L601-L605): when we see a `*` not in the middle of an expr, we add `void =`.

The code then parsing it is [here](https://github.com/gkz/LiveScript/blob/68a6d215e790eccb1047af0b7c91f9b695adadc6/src/grammar.ls#L224-L225), where `Arr.maybe` discards the Array-like structure for a single element.

I could see something like (parsing `*` as `LISTAR`):

```livescript
ListarList:
  o 'LISTAR Expression' -> [$2] # we might want an OptComma here
  o 'ListarList NEWLINE LISTAR Expression' -> $1 ++ $4
  # contrarily to ArgList, I don't think we want a `ListarList NEWLINE DEDENT ListarList DEDENT` rule.

# in Expression:
        o 'Chain ASSIGN INDENT ListarList OptComma DEDENT'
        , -> Assign $1.unwrap!, $4, L 2 Box $2
```

with that, we get:

```
a =
  4
# var a; a = 4;

b =
 * 4
# var b; b = [4];

c =
 * x:
    * 5
# var c; c = [{x: [5]}];
```

rhendric:
At the very least, this should go through a deprecation release before the change is made.

Would you also want to change the related ‘surprise’
```ls
->
  * foo: 0
  * bar: 1
# function(){ ({ foo: 0 }); return { bar: 1 }; }
```
?

vendethiel:
I think the proposed grammar would fix this "surprise".

vendethiel:
Of note: the `ArgList INDENT ArgList DEDENT` rule is mainly for this use case:

```livescript
[ a
  b
  c ]
```
And other similar cases.

rhendric:
Another question is whether
```lsc
[
  * a: 1 b: 2
  * c: 3
]
```
would continue to be a list of objects, or whether you would make it a list of a list of objects.

vendethiel:
> or whether you would make it a list of a list of objects.

That's a good question. Definitely a breaking change if we pick the latter.

It's still possible, though ugly, to separate the implicit objects in this case:

```livescript
[
  a: 1
  b: 2
,
[...]

PR #1001, opened 2018-01-18
RFC: support compiler warnings in a less ad-hoc way

rhendric:
Since #987 got blocked on a compiler warning, but I wanted to write tests that would trigger that warning, I decided we should make compiler warnings more of a thing, largely so they could be disabled in tests (or other situations where they need to be disabled).

This of course opens the gates to more compiler warnings being used in more places where we want to eventually break compatibility or ensure that dumb things don't happen (see: #1000).

However, even though this is a small amount of code, it's adding a compiler flag which I think of as a big change. Want to get lots of feedback on this before I merge it.

### DIFF
--- a/lib/ast.js
+++ b/lib/ast.js
@@ -388,9 +388,17 @@ SourceNode::to-string = (...args) ->
   },
   carp: function(msg, type){
     type == null && (type = SyntaxError);
-    throw type(msg + " on line " + (this.line || this.traverseChildren(function(it){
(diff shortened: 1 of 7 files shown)


vendethiel:
We really need to cleanup the divide between `a or EXPR` and `EXPR unless a` (`and`/`if` as well).
Postfix `unless`-es like this tend to be hard to notice.

I agree with the sentiment, more generally.

PR #1010, opened 2018-02-26
Fixes #1009 Reset lexer before tokenization

### LINKED ISSUE #1009: Cannot reuse same instance of lexer - identifiers property is never reset
I'm using same livescript instance inside  atom editor plugin to preview, autocomplete, lint multiple files.
But when I made mistake in a file I would got an error "Inconsistent use of ...". This error persisted even if after I corrected it, because lexer uses property `identifiers` defined at [line](https://github.com/gkz/LiveScript/blob/0e420e01f3f17e17aba5a1e31607ea29f90e9f85/src/lexer.ls#L111) that is never reset. 
Now I use wrapper for livescript compiler so no more errors for me, but problem in lexer still persists and IMO it should be addressed.

bartosz-m:
Fix for #1009 moved initialization of lexers property `identifiers` and `dend` inside `tokenize` method to ensure that whenever lexer is used multiple times in a row, it starts with fresh state.

### DIFF
--- a/lib/lexer.js
+++ b/lib/lexer.js
@@ -19,6 +19,7 @@ exports.rewrite = function(it){
 };
 exports.tokenize = function(code, o){
   var i, prevIndex, c, charsConsumed, that;
+  this.reset();
   this.inter || (code = code.replace(/[\r\u2028\u2029\uFEFF]/g, ''));
   code = '\n' + code;
   this.tokens = [this.last = ['NEWLINE', '\n', 0, 0]];
@@ -104,6 +105,10 @@ exports.tokenize = function(code, o){
 };
 exports.dent = 0;
 exports.identifiers = {};
+exports.reset = function(){
+  this.dent = 0;
+  this.identifiers = {};
+};
 exports.hasOwn = Object.prototype.hasOwnProperty;
 exports.checkConsistency = function(camel, id){
   if (this.hasOwn.call(this.identifiers, camel) && this.identifiers[camel] !== id) {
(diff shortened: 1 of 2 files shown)


rhendric:
Did you test to make sure this doesn't affect the REPL (`identifiers` should keep its value between lines)?

bartosz-m:
Totally forgot about REPL. Checked now and as suspected `identifiers` are not preserve between lines, so I will find another way to fix this issue.

bartosz-m:
Reset logic lives now inside method that is called only if flag `autoreset` (default: false) is set to `true` so all behavior is preserved but when in need of reusing lexer one can alway set `autoreset` to `true` or manually invoke `reset`

rhendric:
Is `autoreset` really necessary? Anything capable of setting `autoreset` is also capable of calling `reset`.

bartosz-m:
It is not necessary and I wont insist on keeping it but it makes whole solution more I in spirit of DRY - onetime setting of flag vs calling reset in every places that need to reuse lexer.
So remove it?

rhendric:
If you remove it, and add a comment to the effect that `reset` is expected to be called by external systems with use cases similar to yours, I would merge this (after maybe a day for me to think of other possible objections, which I don't really anticipate having).

That's not to say that I'm completely against `autoreset`, in this or perhaps a future PR, but it would take a little more convincing.

(One other minor request: change the bare `@` in the implementation of `reset` to `this`, to be consistent with code style elsewhere in the compiler.)

bartosz-m:
Is that comment  sufficient?

rhendric:
Yes, thank you! If I don't have this merged in 24 hours, feel free to poke me.

PR #959, opened 2017-03-29
remove process.argv mutation

rhendric:
Fix #916. See that issue for commentary.

PR #1014, opened 2018-03-06
fix script argument regression, add process.argv.lsc

### LINKED ISSUE #1013: passing command line arguments to lsc executable still not working?
my apologies if i am being stupid, but i can not manage to pass arguments to a `#!/usr/bin/env lsc` executable on the command line, as `lsc` attempts to read them as additional input files

my expected behavior was for any arguments after the input file to be passed onto the script and available with `process.argv`

i have tried with `--` as suggested in #610, and i have been reading similar issues (#569, #587, #589, etc) to find a solution to no avail

any guidance would be appreciated.

rhendric:
In commit 03c6e6a, removing the code that mutated process.argv also removed a side effect that prevented all arguments to lsc from being handled as scripts to compile and run. This caused a regression when running `lsc script-file.ls arg1 arg2` that would result in a crash when no `arg1.ls` file could be found.

The regression is fixed by reintroducing the side effect to remove all but the first positional argument from the array containing the scripts to compile and run, but only when lsc is running scripts—batch compilation commands involving multiple input files should remain unaffected.

This commit also adds a `process.argv.lsc` property, present only when a script is being interpreted by lsc, which is an array holding the name of the script being invoked along with any extra arguments not consumed by lsc. The intent here is to provide a way to get script arguments that's simpler and more reliable than picking through process.argv and skipping over elements like node, the lsc compiler itself, and any CLI options being provided to either.

---

Fixes #1013. This is both a bug fix and an experimental feature (the `process.argv.lsc` part), so I would like to hear some feedback on whether this mild alternative to the original `process.argv` mutation that I nuked back in #959 is reasonable—if not, it wouldn't be hard to remove that part and just make this the bug fix.


rhendric:
Crickets and thumbs up, so I'm issuing final call for comments on this. Will merge on or after **Friday, March 23** if nobody has anything to say.

corwin-of-amber:
Passing `process.argv.lsc` to command line argument parsers (e.g. `commander`) results in an error since these in fact expect the `node` executable to be the first argument. So I have to do something like:
```
opts.parse (if process.argv.lsc then ['', ...process.argv.lsc] else process.argv)
```

rhendric:
@corwin-of-amber, that's an interesting point. Part of why I wanted to add the `.lsc` property instead of continuing to mutate `process.argv` itself was that I didn't think there was a good way to uniformly adhere to the various expectations that different libraries placed on `process.argv`; this feels like another one of those. But maybe there's little harm in adding a dummy entry to the start of `process.argv.lsc` just so that it can easily be used as a substitute in this particular use case. It strikes me as a hack, but maybe it's worth it?

corwin-of-amber:
Sounds to me like `process.argv.lsc[0]` should be `lsc` itself – as it is the interpreter running the current script (similar to `bash`, `python`, `node` etc.).

Issue #542, opened 2014-08-12
a[]b ++ c fails to compile


quarterto:
`Unexpected ','`

PR #1026, opened 2018-03-26
Fix for #542

pepkin88:
Fix for #542

I saw more unrewound or seemingly incorrectly rewound indices after `splice`s in rewriters, they might possibly be causing some bugs.


rhendric:
Thanks, nice work!

Issue #497, opened 2014-04-28
Introducing "executable" js option


kristianmandrup:
``` LiveScript
# options#L45
    * option: 'exe'
      alias: 'x'
      type: 'Boolean'
      description: 'compile to executable Script'
```

Integration of option in `command.ls`. Not exactly elegant yet. 

``` LiveScript
# command.ls#L148
# Write out a JavaScript source file with the compiled code. By default, files
# are written out in `cwd` as `.js` files with the same name, but the output
# directory can be customized with `--output`.
!function write-JS source, js, base, json
  #     foo.ls     => foo.js
  #     foo.jsm.ls => foo.jsm
  ext = 'X' if o.exe
  ext = '.json' if json
  extension = if ext is void then '.js' else ext
  filename = path.basename(source)replace do
    /(?:(\.\w+)?\.\w+)?$/ -> &1 or ext
  filename = filename.replace /X$/, ''

  dir = path.dirname source
  if o.output
    dir = path.join that, dir.slice if base is '.' then 0 else base.length
  js-path = path.join dir, filename

  if o.exe
    js = "#!/usr/bin/env node\n\n" + js

  !function compile
    e <-! fs.write-file js-path, js || '\n'
    return warn e if e
    util.log "#source => #js-path" if o.watch
    # add executable flag to js file
    if o.exe
      say "executable", js-path
      xe <- fs.chmod js-path, '0755'
      return warn xe if xe
    e <-! fs.stat dir
  return compile! unless e
  require 'child_process' .exec do
    "mkdir #{['-p' unless /^win/test process.platform]} #dir" compile
```

Matthew-Davey:
I was just about to raise an issue for this, glad I searched the archive first! :+1:

PR #1032, opened 2018-03-27
Fix shebang and header ordering

### LINKED ISSUE #1030: Better shebang support needed
Currently
```livescript
``#!/usr/bin/env node``
```
compiles to
```javascript
#!/usr/bin/env node
(function(){

}).call(this);
```
I know there are options that allow to remove `Generated by` line as well as function wrapper, but I believe this should work properly even if those are not used.

Ideally, LiveScript should blindly preserve anything that is on the first line and starts with `#!`, without any need for backticks or additional CLI options.

TL;DR:
```livescript
#!/usr/bin/env node
console.log 'Hello'
```
should compile to
```javascript
#!/usr/bin/env node
(function(){
  console.log('Hello');
}).call(this);
```
by default.

rhendric:
Earlier support for shebang lines in the compiled output moved them before the default function wrapper, but not before the default LiveScript header comment. This moves shebangs above the header comment as well.

Fixes #1030.

---

#1030 seems like a reasonable request to me, but I might be missing some reason not to do this—speak up here if you can think of any. (There's some prior discussion at #497, which mostly just rejects new explicit compiler options in favor of the existing workarounds, but nobody proposed what #1030 proposes.) I'll hold this PR for **two weeks**, merging on or after **April 10** if there are no objections.

**Edit:** It looks like #1030 as written isn't a good idea, so this PR now targets [this comment](/gkz/LiveScript/pull/1032#issuecomment-376808345).


rhendric:
(One possible reason to object: shebangs ought to refer to a file's actual contents, not the contents of a derived file. `#!/usr/bin/env node` is technically incorrect for LS files. Conversely, an LS script can correctly use `#!/usr/bin/env lsc` and be executable, but with this change if that file were also to be compiled, the output would have an incorrect shebang.)

pepkin88:
For files intended to run with `lsc`, shebang should be:
```
#!/usr/bin/env lsc
```
But for files intended to be compiled with `lsc` and then run with `node`, the shebang line should be:
```
``#!/usr/bin/env node``
```
just like in example from #1030.
Why? Because it doesn't introduce any new syntax or exception to a rule.
Moreover, there wouldn't be incorrect shebangs left in files. The first example would strip the comment with an lsc shebang from the compiled file (correct), in the second example ``` ``#! ``` shouldn't be interpreted as a shebang (correct).

Flipping lines can stay, but I would rather test the compiled code for starting from `#!`.

rhendric:
@nazar-pc, what do you think about that? If

```ls
``#!/usr/bin/env node``
console.log 'Hello'
```

compiled by default to

```js
#!/usr/bin/env node
(function(){
  console.log('Hello');
}).call(this);
```
would that look good to you?

nazar-pc:
Having both support for running original script with LiveScript and compiled with Node directly seems highly unlikely, but still possible.

So if I understand correctly
```livescript
#!/usr/bin/env lsc
``#!/usr/bin/env node``
console.log 'Hello'
```
would be still compiled by default to
```javascript
#!/usr/bin/env node
(function(){
  console.log('Hello');
}).call(this);
```
and work properly all the time.

I'm fine with that, makes perfect sense.

nazar-pc:
Found minor issue with this:
```livescript
/**
 * Whatever
 */
```
is compiled to
```javascript
/**
 * Whatever
 */
```
instead of previous
```javascript
/**
 * Whatever
 */
```

GitHub seems to use this header for identifying whether `*.js` file is standalone or was compiled from other language like LiveScript and starts showing incorrect distribution of languages used by project.

Would you be so kind to fix this and add one more test case?

rhendric:
I'll look into it, thanks!

Issue #999, opened 2018-01-13
Typings bikeshedding


vendethiel:
Some time ago (#803) I had started working on a PR to add type hints to the language.

That's still something I miss to this day, so I want to bring it back up.
There were a lot of discussions, so I think we should try to discuss it beforehand.

# Typing symbol

I suggested the symbol `@:` to add a type annotation. Someone suggested `^`. @gkz suggested `::`.

The reason I want(ed) not to use `::` is because it's rewritten to `ID:prototype` when it's parsed directly, at which point we have no info on whether it's going to be part of a signature of just a Parens (we only tag `()` as `PARAM(` and `PARAM)` after seeing `->`).
We'd need to go through the tokens to mark `ID:prototype` as `ASCR` (or whatever else). We'd also need to be careful not to tag in "defaults" – that is, `(x = -> ::) ->`, `(x ? ::) ->`, `(x and ::) ->`, etc...

# where to allow typings

I'd suggest we only allow type annotations in:
 - function declarations (`function` and `->`/`<-`/...)
 - `var`
 - `let`

# generics

I have one suggestion: `[T](xs @: Array[T]) ->`. The reason I suggest `[]` rather than `<>` is mostly for parse-ability (and because I'm used to Scala...). Tagging `[]() ->` is easier than the ambiguities with `<>`, we just have to look for a `]` when re-tagging a `(` as a `PARAM(`, and cycle back to `[` from there (since an array is never callable anyway).
(since `[T]() ->` and `xs[T]() ->` make no sense anyway)

Parsing the `@: Array[T]` part is harder, but still simpler than `@: Array<T>`, where the `>` would be parsed as a BIOP because it's directly followed by a `)`...

--------------

@rhendric I'm interested in your PoV, because it's a sensible topic, considering all the ambiguities to keep in mind :-).

vendethiel:
As is the tradition, I've first implemented this with **~~ LEXER HACKS~~**: 509cf13390336068679c30c4b36f53607bbff849

```
$ ./bin/lsc -m none -bce '(x @: Map[Int, Array[Int]], y @: Int) ->'
(function(x : Map<Int,Array<Int>>, y : Int){});
```

rhendric:
[...]

PR #1033, opened 2018-03-27
Remove `where` from reserved keywords

pepkin88:
As documentation says, `where` feature was deprecated in v1.1, removed in v1.2.

Unless there are some plans for `where` as a global keyword, I think it should be removed from the reserved word list.


vendethiel:
I know we aren't many people in the camp "bring back `where`", but... I do like the feature.
[Especially as discussed in #999](https://github.com/gkz/LiveScript/issues/999#issuecomment-359178336).

PR #1036, opened 2018-03-29
Fix #1035

### LINKED ISSUE #1035: `for..else` doesn't work with the implicit `for array` variant
Compiling this:
```LiveScript
for [1 2 3]
  ..
else 0
```
gives an error `Parse error on line 3: Unexpected 'ELSE'`.

Working examples:
```LiveScript
for n in [1 2 3]
  n
else 0
```
```LiveScript
for [1 2 3] when 1
  ..
else 0
```

rhendric:
This is just a quick grammar tweak; honestly I'm shocked that this hasn't been caught and fixed earlier. But here we are.

Fixes #1035. Will merge on or after **April 5**.


summivox:
Oops I didn't knew LSC had a `for else` in the first place... guess that's why it had not been caught!

PR #1037, opened 2018-03-29
Suppress the strict comparison warning when comparing `..` with a literal

pepkin88:
Currently, a code like
```LiveScript
1
  .. == 1
  1 == ..
```
will produce misleading warnings upon compilation:
```
WARNING: strict comparison of two different types will always be false: .. == 1
WARNING: strict comparison of two different types will always be false: 1 == ..
```

It does that because `..` is interpreted as a callable literal, probably for compiler's simplicity. That's why I added a simple exception to the conditions when to show the warning. I also have checked for other symbols requiring attention, but I think it's only `..` (I checked `Literal::compile`'s switch cases and the `Literal` class in general).

I don't know how to test for warnings, to I left the commit without tests.


vendethiel:
Most probably `..` shouldn’t be a literal...

vendethiel:
Would you mind going the extra mile, and instead [not represent `..` as a Literal](https://github.com/gkz/LiveScript/blob/master/src/lexer.ls#L525)?
There [shouldn't be](https://github.com/gkz/LiveScript/blob/master/src/ast.ls#L568-L570) [that much code](https://github.com/gkz/LiveScript/blob/master/src/ast.ls#L2049-L2050) [to change](https://github.com/gkz/LiveScript/blob/master/src/ast.ls#L1721)

pepkin88:
I'll look into that.
For now, I'll leave it there. There's no rush for that fix :)

vendethiel:
FWIW the issue is also visible with `1 == this`.

rhendric:
I suppose `this` ought not to be a `Literal` either?

vendethiel:
Would fix a few other issues, yes.

rhendric:
Oh, neat. Which ones did you have in mind?

vendethiel:
```~> `this`;``` too I think.

rhendric:
(Assuming you meant double-backticks there, and GitHub ate them; if not, I'm very confused!)

Hmm, I don't see it. That `this` will end up inside a `JS` node, regardless of what kind of node a bare `this` is. To make that behave the same as `~> this`, I think you'd have to change `Fun::uses-this` to parse the contents of `JS` children to see if they contain `this`. (Which doesn't sound worth doing to me, TBH.)

vendethiel:
Oops, that's on me, I totally forgot `JS` solved those issues already. (and yes, I did not miss infix this)

PR #1041, opened 2018-04-02
lexer: handle words literals with more care

### LINKED ISSUE #739: can not compile empty list of words

rhendric:
This fixes three related issues around words literals (`<[ ]>`) being
mistaken for newlines or operators when their contents had the same
values as those other token types.

Fixes #739, closes #808.

---

All but the most trivial tweaks to the lexer make me nervous. Will hold for **two weeks**, merging on or after **April 16** if no feedback.

### DIFF
--- a/test/literal.ls
+++ b/test/literal.ls
@@ -119,6 +119,25 @@ compile-throws 'unterminated words' 5 '''
   <[
   '''
 
+# [LiveScript#739](https://github.com/gkz/LiveScript/issues/739)
+# For the below tests, it's important that there not be any extra whitespace
+# between the <[ ]> brackets. These tests guard against mistaking a WORDS token
+# for some other kind of token based on its value only.
+
+# Don't interpret WORDS:\n as a newline
+w = <[
+]>
+eq w.length, 0
+
+# And don't interpret WORDS:++ as concat
+w = <[++]>
+eq w.0, \++
+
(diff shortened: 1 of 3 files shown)


test/literal.ls
```diff
@@ -119,6 +119,20 @@ compile-throws 'unterminated words' 5 '''
...
+]>
+eq w.length, 0
+
+# And don't interpret WORDS:++ as concat
+w = <[++]>
+eq w.0, \++
+
+# And don't interpret WORDS:. as compose
+# (Note: the spacing and punctuation in this test is significant)
+eq '' + <[.]> \.
```
summivox: What happens if you put whitespace around `.` in the word list

rhendric: If there's whitespace on either side of the `.`, the issue isn't triggered (before and after this patch). The lexer packs everything, whitespace included, between the brackets into the `WORDS` token, and only got confused when the value of that token exactly matched some expected values of other types of token.

summivox: Consider putting this explanation as comment in test (explaining why the spacing is significant)?

PR #1050, opened 2018-04-13
fix regression caused by 41fb1a8

### LINKED ISSUE #1030: Better shebang support needed

rhendric:
That attempt to move shebang lines above the header comment also moved
initial source comments above the header comment, which was unintended.
If any of these three elements are present, the correct order is:

  * shebang
  * Generated by LiveScript header comment
  * source file header comment

---

See [report in #1032](/gkz/LiveScript/pull/1032#issuecomment-381212282). Closes #1030 (again).

Will hold for **one week**, merging on or after **April 20** if there are no objections.


nazar-pc:
Works good for me locally, thanks!

Issue #1025, opened 2018-03-26
Expressions starting from `_` are treated as placeholders in `match` cases


pepkin88:
Example:
```LiveScript
match a
| _.isString => 1
| (_.isString) => 1
```

Result:
```js
var ref$;
switch (ref$ = [a], false) {
case !true:
  1;
  break;
case !_.isString(ref$[0]):
  1;
}
```

Expected result:
```js
var ref$;
switch (ref$ = [a], false) {
case !_.isString(ref$[0]):
  1;
  break;
case !_.isString(ref$[0]):
  1;
}
```

The example shows, that wrapping the expression in parentheses is enough to bypass this bug.

rhendric:
Oh, `match`. The undocumented, largely untested, experimental feature that somehow people keep discovering anyway.

This sure does look like a bug, but I don't know whether this should be an error, your expected result, or something else. `_` has special meaning inside `match`, which is complicated by the fact that `_` is also used to create a partial-application function and as far as I know, every case of a `match` expression should be interpreted as a function (unless it isn't, like with constants).

I don't really intend to do anything to improve the state of `match`, unless someone feels like fully and properly specifying its behavior. The good news is, as an experimental undocumented feature, I also wouldn't feel even a tiny bit bad about ripping it out and replacing it with something else. You were all warned. :slightly_smiling_face:

pepkin88:
OK, fair enough.

I drew an analogy to the partial application or switch's default, where only a standalone underscore has a special meaning. In those cases using `_.toString` wouldn't trigger anything special.

I might fix it later. For now, let this issue be a walk-around suggestion.

pepkin88:
Related issue with a plausible use case scenario:

This:
```LiveScript
(_.isPlainObject || _.isArray) value
```
compiles to:
```js
true || true;
```
instead of:
```js
_.isPlainObject(value) || _.isArray(value);
```

Although I think this feature is a real gotcha and makes sense only in the context of `match`, it actually is documented.

rhendric:
Ha, okay, you got me! This should be fixed.

pepkin88:
This looks like an easy one I could tackle, but only from next week.

rhendric:
I'll hold it for you then.

anant-svc:
Hi, tried to get expected result @pepkin88 needed using below:
```
u = _
match a
| u.isString => 1
| (u.isString) => 1
```

Result:
```
var u, ref$;
u = _;
[...]

PR #1059, opened 2018-07-29
Fix for #1025

pepkin88:
Fix for #1025.

This fixes placeholder handling in `match` and with calling logic operators, although the latter doesn't need placeholders at all.

I still think this calling logic operators feature is a miss. In terms of usability and intuition, I strongly prefer `match` over this, even though `match` isn't even documented.

### DIFF
--- a/lib/ast.js
+++ b/lib/ast.js
@@ -1095,7 +1095,9 @@ exports.Chain = Chain = (function(superclass){
   };
   Chain.prototype.autoCompare = function(target){
     var test;
-    test = this.head;
+    if (!this.tails.length) {
+      test = this.head;
+    }
     switch (false) {
     case !(test instanceof Literal):
       return Binary('===', test, target[0]);
--- a/src/ast.ls
+++ b/src/ast.ls
@@ -708,7 +708,7 @@ class exports.Chain extends Node
         this
 
     auto-compare: (target) ->
-        test = @head
+        test = @head unless @tails.length
         switch
         | test instanceof Literal
             Binary \===  test, target.0
--- a/test/operator.ls
+++ b/test/operator.ls
@@ -988,3 +988,15 @@ ok not (even or 1) 3
 ok ((.length > 4) or [1 2 3]) [1 2 3]
 
 eq 8 ((-> &0 + &1 is 5) and (**)) 2 3
+
+# [LiveScript#1025](https://github.com/gkz/LiveScript/issues/1025)
+# The placeholder logic from `match` is also applied here
+_ = -> -> false
+_.isString = -> typeof it == \string
+_.isNumber = -> typeof it == \number
+eq false (_.isString || _.isNumber) []
+eq false (_.isString || _!) []
+eq true  (_.isString || _.isNumber) 1
+
+# other examples of Chains with tails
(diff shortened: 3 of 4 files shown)


rhendric:
Thanks for contributing! Is there a reason this added check only applies to the `_` clause? It seems to me like it should guard `test` being assigned or something.

Also, please squash this down to a single commit.

rhendric:
Also also, there should be no change to `parser.js`; I assume the reason you have a diff there is because you're running Node.js <10? If you can't/don't wish to upgrade to Node.js 10, just don't `git add` that file to the commit.

pepkin88:
From what I understand, the `| test instanceof Var and test.value is \_` line was there to detect the use of a placeholder `_`, as `_` outside of `match` is a variable. The problem was, this test was too broad and caught not only `_`, but also `_.isString` (and calls too, like `_!`, as I just noticed, I'll add a test for that in a moment). So I narrowed the test by checking if the chain doesn't have any tails (like in the case of `_.isString`). The other tests don't need such narrowing. That is what you are asking, correct?

Regarding squashing, isn't it done while merging, as they said [there](https://blog.github.com/2016-04-01-squash-your-commits/)? Other than that, I don't quite know how to do that, I don't understand why those `Merge remote-tracking branch 'gkz/master'`s pop up there. I'm using GitKraken UI for this and it looks like it's ok, but I'm really not sure, I'm not a git expert.

Regarding the changes in `parser.js`, that's correct, I haven't set up Node.js 10 yet, I'll upgrade it.

rhendric:
Everything you've observed is correct, but I think you've only gotten 90% of the way to the root of the problem. The *entire* `auto-compare` method only examines the head of the chain, and if a head matching a special case is found, the existence of any tails is ignored (and any such tails, if present, are discarded). You've fixed this problem for only one case in that method; consider what might happen with, for example,
```ls
match option
| 'widgets'starts-with => widgets
| 'gadgets'starts-with => gadgets
```

I think the better fix would be to examine `@head` only if `@tails` is empty, like `test = @head unless @tails.length`, thus sending all tail-having chains to the `otherwise` clause. What do you think?

Re Git strategy, it's true that I could squash the commits myself (through GitHub or through Git), but it's nicer if contributions come organized in the commit(s) they'll have in the repository (that way, the original author can write the commit messages and the maintainer doesn't have to rewrite anything). While there's no universally agreed upon standard for commit organization, the standard I favor is one commit per feature or fix, which *usually* means one commit per pull request but can sometimes mean a few if there are multiple, interrelated features being developed and reviewed together. I don't know much about Git GUIs, but the easiest way to squash down your commits from the command line is `git rebase -i origin/master`, and change all commits but the first from `pick` to `squash` (or `s`). Then `git push -f` to update the pull request (pull requests are an exception to the rule not to force-push to remote repositories).

pepkin88:
I think I know now what I do wrong with those commits, the next PR has only one. But this one will probably have to be squashed while merging to master, sorry about that.

pepkin88:
Oh, I got you now, you are right. Good test case.

pepkin88:
I stumbled upon next bug, looking for a way to fix this too.

pepkin88:
I think it's good now :) Take a look

rhendric:
Great, thank you!

Issue #1023, opened 2018-03-23
Loop guards (`when`, `case`, `|`) don't work with `for..let` loops with `yield` (or `async`) in their bodies


pepkin88:
Example:
```LiveScript
fn = ->*
  obj = a: 1
  for own let k, v of obj when v % 2
    yield v
```

Result:
```js
var fn;
fn = function*(){
  var obj, i$, k, v, own$ = {}.hasOwnProperty, results$ = [], fn$ = function*(k, v){
    return (yield v);
  };
  obj = {
    a: 1
  };
  for (i$ in obj) if (own$.call(obj, i$)) {
    results$.push((yield* (k = i$, v = obj[i$], v % 2 ? (fn$.call(this, k, v)) : void 8)));
  }
  return results$;
};
```

Expected result:
```js
var fn;
fn = function*(){
  var obj, i$, k, v, own$ = {}.hasOwnProperty, results$ = [], fn$ = function*(k, v){
    return (yield v);
  };
  obj = {
    a: 1
  };
  for (i$ in obj) if (own$.call(obj, i$)) {
    if (obj[i$] % 2) {
      results$.push((yield* (k = i$, v = obj[i$], v % 2 ? (fn$.call(this, k, v)) : void 8)));
    }
  }
  return results$;
};
```

I chose an example with `own`, because it also inserts its own kind of guard and it seems to be working correctly. A guard for a loop with `async` doesn't work either, but I think it's related to #1021.

Other `for..let` variants seem to be OK, those with an else block too.

rhendric:
Also another effect of #1019.

pepkin88:
Now I noticed, that that code doesn't work, but from different reasons. The guard `v % 2` is indeed inserted, but in the wrong spot. But, as you predicted, fix for #1019 (#1060) fixes it too.

Issue #1021, opened 2018-03-22
`await` doesn't work in `for..let` loops


pepkin88:
Example:
```LiveScript
fn = ->>
  for let v in [Promise.resolve 1]
    await v
```

Result:
```js
var fn;
fn = async function(){
  var i$, ref$, len$, results$ = [], fn$ = function*(v){
    return (await v);
  };
  for (i$ = 0, len$ = (ref$ = [Promise.resolve(1)]).length; i$ < len$; ++i$) {
    results$.push((yield* (fn$.call(this, ref$[i$]))));
  }
  return results$;
};
```

Expected result:
```js
var fn;
fn = async function(){
  var i$, ref$, len$, results$ = [], fn$ = async function(v){
    return (await v);
  };
  for (i$ = 0, len$ = (ref$ = [Promise.resolve(1)]).length; i$ < len$; ++i$) {
    results$.push((await (fn$.call(this, ref$[i$]))));
  }
  return results$;
};
```

It seems the mechanism for `yield`s is also used for `await`s, but those require a special treatment.

I also noticed that asynchronous generator functions are not yet supported, so I couldn't test for that.

rhendric:
Good news: fixing #1019 will also fix this! (At least, the proof of concept fix I wrote for #1019 to make sure that it would be a good tutorial bug fixes this.)

Issue #1019, opened 2018-03-21
`yield` and `await` don't work in `let` blocks


pepkin88:
### `yield`

Example:
```livescript
fn = ->*
  let a = 1
    yield a
```

Result:
```js
var fn;
fn = function*(){
  return (function(a){
    return (yield a);
  }.call(this, 1));
};
```

Expected result:
```js
var fn;
fn = function*(){
  return (yield* (function*(a){
    return (yield a);
  }.call(this, 1)));
};
```

### `await`

Example:
```livescript
fn = ->>
  let a = Promise.resolve 1
    await a
```

Result:
```js
var fn;
fn = async function(){
  return (function(a){
    return (await a);
  }.call(this, Promise.resolve(1)));
};
```

Expected result:
```js
var fn;
fn = async function(){
  return await (async function(a){
    return await a;
  }.call(this, Promise.resolve(1)));
};
```
(AFAIK, unlike `yield`, `await` doesn't need wrapping in parentheses, `!await something` works fine)

---

Sidenote:

This is the first issue of many I'm going to post. Way back I announced, that I am planning to start posting some pull requests for the bugs I've found (https://github.com/gkz/LiveScript/issues/903#issuecomment-242090855), but ended up posting nothing. I even was waiting so long, before I started doing something to fix https://github.com/gkz/LiveScript/issues/1005, despite knowing about this bug for years.

So this time I decided to dump my list of bugs and feature proposals one at the time, at least once a day. I hope I'll find some time and will to learn the internals to help more, by making pull requests. But "done is better than perfect". I have procrastinated for so long and I see, that LiveScript needs a help (with bugfixes, with tools, with promotion, with rewriting), and it hurts me, that I just wait and do nothing about it.

Feel free to post suggestions about how I can be more helpful, given low time resources.
I tried learning how the compiler works, but it's quite complicated and requires more analysis, so I've postponed it. Do you know some good resources to learn how to get into such code?

And I'm sorry for any duplicates. I'm not familiar with the whole issue pool, it's quite daunting.

And don't wait for me with 1.6.0. These all can be included in future versions.

pepkin88:
Regarding the issue, `let`s also don't work with `await` (I'm new to this in LS):
```livescript
fn = ->>
  let a = Promise.resolve 1
    await a
```

Result:
```js
var fn;
fn = async function(){
  return (function(a){
    return (await a);
  }.call(this, Promise.resolve(1)));
};
```

Expected result:
```js
var fn;
[...]

PR #1060, opened 2018-07-29
Fixes bugs with `yield`s and `await`s inside various `let` blocks

pepkin88:
Fix for #1019, #1021 and #1023.

This wasn't that hard as I imagined. Did I get it right?
I also noticed a similar behavior is duplicated in `Node::compile-closure`. Perhaps it could be rewritten to use `Call.let` inside.
I still don't know how everything works, but more and more I get the hang of it.

### DIFF
--- a/src/ast.ls
+++ b/src/ast.ls
@@ -1001,7 +1001,15 @@ class exports.Call extends Node
             ++index
         node <<< back: (args[index] = fun)body
 
-    @let = (args, body, generator = false) ->
+    @let = (args, body) ->
+        has-yield = false
+        has-await = false
+        body.traverse-children (child) ->
+            if child instanceof Yield
+                switch child.op
+                | \yield \yieldfrom => has-yield := true
+                | \await            => has-await := true
+            return true if has-yield and has-await
         params = for a, i in args
             if a.op is \= and not a.logic and a.right
                 args[i] = that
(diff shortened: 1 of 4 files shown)


src/ast.ls
```diff
@@ -1001,15 +1001,26 @@ class exports.Call extends Node
...
-    @let = (args, body, generator = false) ->
+    @let = (args, body) ->
+        has-yield = false
+        has-await = false
+        body.traverse-children (child) ->
+            if child instanceof Yield
+                switch child.op
+                | \yield \yieldfrom => has-yield := true
+                | \await            => has-await := true
+            return true if has-yield and has-await
```
rhendric: A function can never legally have both a yield and an await, so this is a little misleading. Could you change this back to something a little closer to the original expression in `add-body`, where you return as soon as any instance of `Yield` is found? Instead of returning true you can return `child.op`, to distinguish between the yield and await cases without mutating a variable in the outer scope.

pepkin88: This was left intentionally for async generators, as I would imagine they will come to LS eventually. I thought it would be better to future-proof it, in case someone implementing async generators forgets about `async`s and `yield`s in `let` blocks, as it have happened with async functions.

rhendric: That's a good thought! I wasn't aware that async generators had gotten this far in the standardization process before now; they should be very easy to add to LiveScript, so I agree this is a good precaution.

PR #1061, opened 2018-07-30
Feature detection in the test environment

pepkin88:
Currently it was easy to miss testing generators, because of omitting the `--harmony` flag; even `make full` skipped testing generators.
This PR introduces detecting support for certain syntax oriented language features, before running tests.

This doesn't mean the tests will always run/pass on the older Node versions, as sometimes other new features are used, which are not syntax oriented, such as `String::startsWith`. The purpose of this PR is however to be able to always rely on the default `npm test` or `make full` and be sure, that having the newest runtime environment set up, all the tests will run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ browser/livescript-min.js:  browser/livescript.js
 package.json: package.json.ls
 	$(LSC) --compile package.json.ls
 
-.PHONY: build build-browser force full install dev-install test test-harmony coverage loc clean
+.PHONY: build build-browser force full install dev-install test coverage loc clean
 
 all: build
 
@@ -50,9 +50,6 @@ dev-install: package.json
 
 test: build
 	./scripts/test
-
-test-harmony: build
-	node --harmony ./scripts/test
 
 coverage: build
 	$(ISTANBUL) cover ./scripts/test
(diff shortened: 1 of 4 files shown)


pepkin88:
I left all the harmony related scrips/commands intact to preserve backwards compatibility. Although using them will be now obsolete.

rhendric:
I think we can be more aggressive than this. I'm comfortable saying that Node.js 4 is the earliest Node we support running tests on, so you can remove the generator check entirely. I would also support ripping out the test-harmony infrastructure; I don't care about backwards compatibility for people running LS tests. There aren't that many of us, after all.

pepkin88:
I noticed that tests are buggy for Node versions <5.10, because of a different implementation of `Buffer.from` function, which exists, but can only process typed arrays.

pepkin88:
It affects not only tests, but compilation in general, when compiling with embedded source maps.

pepkin88:
I have tested on v4.0.0 and v5.0.0 and those gave me errors. Version 4.9.1 indeed has the newer implementation.

It seems that checking for existence of `Buffer.alloc` is enough to ensure, that `Buffer.from` has the modern implementation.

pepkin88:
I addressed this issue in another PR: #1062.

PR #1062, opened 2018-07-30
Ensure using newer implementation of `Buffer.from`

pepkin88:
Prior to Node v5.10 (and v4.5) it's impossible to use `Buffer.from` for creating buffers from string.
Existence of `Buffer.alloc` implies the newer implementation of `Buffer.from` and possibility to use it with strings.


rhendric:
Thanks for sleuthing this out!

PR #1063, opened 2018-07-30
support async generators with ->>*

rhendric:
@pepkin88 [brought to my attention](/gkz/LiveScript/pull/1060#discussion_r206048053) that async generators are coming down the JS pipe (and already supported by default in Node.js 10.7), and I think they're pretty easy to add at this point and I see no reason to wait. I [said](/gkz/LiveScript/issues/1016) I wouldn't prioritize more features for 1.6 but this is a pretty minor and obvious combination of two existing features, so I'm giving myself an exception. :smile: Open for comments for **two weeks**, to be merged on **August 13** if there are no opposing voices.

@gkovacs, @summivox, as sponsors of the original async work, if you think I've missed anything here I'm interested in hearing about it.


pepkin88:
Could be more tests from generators.

For example:
```ls
# yield from in let
first = ->>*
  i = 0
  loop => yield await Promise.resolve i++
second = ->>*
  let
    await Promise.resolve 1
    yield from first!
list = second!
for let i to 3
  {value} <- list.next!then
  eq value, i
```
(a modified test from `generators.ls` with an addition of a let block)
This passes, but only due to recent changes.

However this one doesn't:
```ls
# in switch
f7 = (x) ->*
    y = switch x
    | true => yield await Promise.resolve 1
    | _    => yield await Promise.resolve 2
    y
g7 = f7 true
g7.next!then -> eq 1 it.value
```
(modified test from `generators.ls`)

Speaking of which, what do you think about my idea from https://github.com/gkz/LiveScript/pull/1060#issue-204653754 to rewrite `Node::compile-closure` to use `Call.let`? That way we would avoid duplicating a similar behavior.

rhendric:
There is a test with a `let` block already in there (which indeed does fail if I revert your fix). I added a test to exercise `yield from`, but I don't think it's necessarily valuable to repeat all of the generator.ls tests in async form.

Re your `compile-closure` proposal, if you want to go forward with that, I'd review it. Maybe the two functions should be unified, but at present both do optimizations that the other does not (`Call.let` always produces a `.call` invocation, whereas `Node::compile-closure` only does so if `this` is used). So I don't think it's a clean case of one should use the other, but if you want to try merging them, be my guest.

summivox:
LGTM

pepkin88:
@rhendric Should I do it on your branch or on master?

rhendric:
Master, unless you expect it to depend on this work for some reason.

pepkin88:
After some thought I think I'll postpone that indefinitely - the gain isn't that big and `Call.let` isn't as compatible as I thought, it would require some custom handling of `arguments`. So I'll leave it for now.

And regarding async generators, looks great 👍

PR #1065, opened 2018-07-31
Fix `--no-warn` flag

### LINKED ISSUE #1064: `--no-warn` flag doesn't work
The command `lsc --no-warn` prints this:
```
Invalid option '--warn' - perhaps you meant '--bare'?
```

Optionator has a special logic for [options starting with `no-`](https://github.com/gkz/optionator/blob/0.8.1/src/index.ls#L276). In this case it tries to set a flag `warn` as false, but `lsc` offers no support for such flag.

Should we change the flag to something else or issue a bug/PR to the Optionator?

pepkin88:
Fixes #1064.

Should I bump the Optionator to 0.8.2 here, or version bumps will be done later?


rhendric:
Wow, I really goofed here, huh? Thanks for catching and fixing this!

There's no relevant difference between 0.8.2 and 0.8.1 here, is there? It looks like you're just using the same pattern that was already in use for `--no-headers`.

pepkin88:
v0.8.2 fixes https://github.com/gkz/optionator/issues/18
npm tries to install the newest compatible version (thus 0.8.2), but if 0.8.1 is already found in node_modules, it leaves it as is.

Issue #897, opened 2016-06-05
lsc is not opening repl


Announcement:
on Windows 7 powershell (haven't tested any other environment yet with livescript 1.5.0) simply typing `lsc` isn't opening a repl.

anyone experiencing the same issue or know what that's about?

dk00:
if it said `TypeError: Path must be a string. Received undefined`
add environment variable 'HOME' with value %USERPROFILE% and it should open.

https://github.com/gkz/LiveScript/pull/695/commits/b32276a07f8fda41b0ee0f8ee75bccf2d549349e#diff-62cab7f67fcaea6ddbe4abb9d1d32828R254
Windows does not have 'HOME', 'HOMEPATH' should be used instead

atifaziz:
> anyone experiencing the same issue or know what that's about?

Yes, with 1.5, I'm seeing the same.

```
B:\>npm i livescript
livescript@1.5.0 node_modules\livescript
├── prelude-ls@1.1.2
├── source-map@0.5.6
└── optionator@0.8.1 (fast-levenshtein@1.1.4, type-check@0.3.2, deep-is@0.1.3, wordwrap@1.0.0, levn@0.3.0)

B:\>node_modules\.bin\lsc.cmd
LiveScript 1.5.0 - use 'lsc --help' for more information
path.js:233
      throw new TypeError('Arguments to path.join must be strings');
            ^
TypeError: Arguments to path.join must be strings
    at Object.win32.join (path.js:233:13)
    at repl (B:\node_modules\livescript\lib\command.js:394:24)
    at B:\node_modules\livescript\lib\command.js:131:7
    at Object.<anonymous> (B:\node_modules\livescript\bin\lsc:7:26)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
```

But using 1.4, I get the REPL with the same steps:

```
B:\>npm i livescript@1.4.0
livescript@1.4.0 node_modules\livescript
├── prelude-ls@1.1.2
├── optionator@0.6.0 (fast-levenshtein@1.0.7, deep-is@0.1.3, type-check@0.3.2, levn@0.2.5, wordwrap@0.0.3)
└── source-map@0.3.0 (amdefine@1.0.0)

B:\>node_modules\.bin\lsc.cmd
[...]

PR #1066, opened 2018-07-31
Fix for #897

pepkin88:
Fix for #897

`os.homedir` was added in Node v2.3.0, fallback values of `process.env.HOME` and `process.env.USERPROFILE` works for POSIX and Windows


rhendric:
This looks good to me, but I'd like a second person to test this on Windows (I assume you did?) just to be sure (and to ensure there are no lurking second bugs hiding behind the one you fixed).

vendethiel:
Works on my windows

pepkin88:
Yes, I checked both `os.homedir!` and `process.env.USERPROFILE`. Windows is my main platform, although I do tests and builds on a virtual Ubuntu machine.

PR #1086, opened 2019-09-26
test

determin1st:
probably faster :/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 t.*
 coverage
 node_modules
+*.swp
+*.vim
+src/*.js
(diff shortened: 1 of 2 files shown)


vendethiel:
Probably?

rhendric:
> probably faster :/

I don't see why it would be, unless you have data to show that it is.

determin1st:
well, i've tested it: https://jsperf.com/object-spread-vs-polyfill/3

oookay, at least, it's not slower... 8) 
theoretically it should be faster as it doesnt create additional variable..

vendethiel:
I think your code gets optimized away.

rhendric:
This is way too speculative an optimization to be worth thinking about. I'm all for contributions that make LiveScript faster, but not PRs that could maybe theoretically make it faster but that in practice have no measurable effect.

PR #1103, opened 2020-06-09
Fix #1102, and other soaking-related fixes

### LINKED ISSUE #1102: Existence check is ignored before object slice access
All of these
```livescript
(.{foo, bar, baz})
(?{foo, bar, baz})
(?.{foo, bar, baz})
-> it{foo, bar, baz}
-> it.{foo, bar, baz}
-> it?{foo, bar, baz}
-> it?.{foo, bar, baz}
```
produce exactly the same output:
```javascript
(function(it){
  return {
    foo: it.foo,
    bar: it.bar,
    baz: it.baz
  };
});
```

Similarly, every single one of these
```livescript
(.<[foo bar baz]>)
(?<[foo bar baz]>)
(?.<[foo bar baz]>)
-> it<[foo bar baz]>
-> it.<[foo bar baz]>
-> it?<[foo bar baz]>
-> it?.<[foo bar baz]>
(.[\foo, \bar, \baz])
(?[\foo, \bar, \baz])
(?.[\foo, \bar, \baz])
-> it[\foo, \bar, \baz]
-> it.[\foo, \bar, \baz]
-> it?[\foo, \bar, \baz]
-> it?.[\foo, \bar, \baz]
```
also produce identical output:
```javascript
(function(it){
  return [it['foo'], it['bar'], it['baz']];
});
```

The expected behaviour here, naturally, would be to make an additional nil check when `?` is used (or at least produce a compile error if such syntax is not supported).

rhendric:
This PR contains five bug fixes (two of which are regressions that I caused), a very minor feature, and some housekeeping, all related to or needed by the interactions between `rewrite-shorthand` and `unfold-soak`. The title commit reads as follows:

---

This issue was a regression from 1.5.0 first appearing in commit
8cffc57, in which I introduced the `rewrite-shorthand` method and moved
slice expanding into it. The problem with my refactoring is that it
caused everything in `rewrite-shorthand` to happen before `unfold-soak`,
but the code in the issue relies on `unfold-soak` happening prior to
slices being expanded.

This commit prepends a short-circuiting call to `unfold-soak` in the
`rewrite-shorthand` method of every node type that defines a non-trivial
`unfold-soak`.

---

It seems I made quite the mess when introducing `rewrite-shorthand`! In addition to the regressions tested and fixed here, there are a few other shorthand-rewriting regressions which will be fixed in the next PR. This PR was getting quite large though, so let's get this merged first.

These fixes aren't especially complicated, but I'm feeling especially humbled by how much I screwed up, so I'm going to set the review period for this PR at **three weeks**; I will merge on or after **June 29** if there are no comments.

Fixes #1102.


determin1st:
im sure you'll make it! add more comments into code for the next readers, who knows what happen to github

rhendric:
@determin1st, anything in particular that doesn't seem clear to you? All of the new code reads as fairly self-explanatory to me, but getting other perspectives on that sort of thing is why I put myself through review.

PR #1104, opened 2020-06-29
fix #809

### LINKED ISSUE #809: "that" should be aware of —const
So it doesn’t try and clobber itself behind the scenes when used multiple times, breaking at compile time

bung87:


rhendric:
Hi, thanks for contributing! Could you please:
 * Add tests in the appropriate file (for this, I think `test/declaration.ls`)
 * Run `make` and add the `lib` directory to the commit

Please make these (and any future) changes by amending the existing commit and force pushing, not by adding new commits.

Thank you again!

bung87:
Hi, I followed you suggestions, think things get done.

PR #1105, opened 2020-06-30
Fix rewrite-shorthand regressions

rhendric:
As a continuation of the mea culpas in #1103, here are some more things that broke with my rewrite-shorthand refactoring.

This is a simpler pack of fixes than the last one, so I'm giving it **two weeks**, to be merged on or after **July 14** if there are no comments.

I'd also like to attempt a point release after this and possibly #1104 are merged, since there are now a substantial number of regressions associated with the previous release which will be fixed.

### DIFF
--- a/src/ast.ls
+++ b/src/ast.ls
@@ -932,8 +932,9 @@ class exports.Chain extends Node
     rewrite-shorthand: (o, assign) ->
         return that.rewrite-shorthand o, assign or that if @unfold-soak o
         @head = that if @head.rewrite-shorthand o
+        last-i = @tails.length - 1
         for item, i in @tails
-            @tails[i] = that if item.rewrite-shorthand o, assign
+            @tails[i] = that if item.rewrite-shorthand(o, assign && i is last-i)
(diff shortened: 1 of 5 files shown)


src/ast.ls
```diff
@@ -932,8 +932,9 @@ class exports.Chain extends Node
     rewrite-shorthand: (o, assign) ->
         return that.rewrite-shorthand o, assign or that if @unfold-soak o
         @head = that if @head.rewrite-shorthand o
+        last-i = @tails.length - 1
         for item, i in @tails
-            @tails[i] = that if item.rewrite-shorthand o, assign
+            @tails[i] = that if item.rewrite-shorthand o, assign && i is last-i
```
vendethiel, line 937: What’s `that` going to be with that `if`?
Why keep the loop?

rhendric: Perhaps it would have been clearer if I wrote `@tails[i] = that if item.rewrite-shorthand(o, assign && i is last-i)` ?

PR #1118, opened 2022-01-01
enlarge gihub-stars iframe due to numbers over 2k

dym-sh:
current:
current_20220101143418

fixed:
should_20220101143142


anko:
Is it possible to make an iframe dynamically size to fit content?

Could use an SVG from [Shields.io](https://shields.io/) to avoid the iframe completely.  They can dynamically generate buttons from an API lookup on their backend, based on URL parameters like this:

```
https://img.shields.io/github/stars/gkz/LiveScript?style=social
```
[Principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) and all that.

dym-sh:
the reason it uses iframe currently is b/c its "serverless" – request is made by a js-script to github api at the moment of loading

i'd rather ask about the entire point of having dynamic counter in the first place, past 2k stars it doesnt really matter, can just get the number on each release as a ci/cd step

PR #1129, opened 2024-03-05
A live command, like the cake in CoffeeScript

othelarian:
# Origin of the PR

This PR is more like a fun addition to LiveScript. I always wonder why an equivalent to the CoffeeScript `cake` command doesn't exist un LiveScript, with a `Livefile` file at the package's root. So I did it.

# What it does

I added `scripts/build-live` to compile the `bin-src/live.ls` script into `bin/live`, and also a `Livefile` as an example of what is possible.

It works like this:

- `option [name], [alias], [type], [description]` => like in a `Cakefile` but with type, to make it work with `optionator`
- `task [name], [description], [callback]` => also like in a `Cakefile`, the exact same syntax

Also, I added a new target in the `makefile`, `live: bin/live`, to run the script that compile `bin-src/live.ls` into `bin/live`.


ceremcem:
Can we use this in place of Gulp?

othelarian:
> Can we use this in place of Gulp?

Completely, it is what I do for all my project, but in CoffeeScript. I also use `bach`, which is a lib `gulp` uses to create the pipeline under the hood. If you want the `gulp` pipelines you just have to add `bach` and `vinyl`.

